PR #44, opened 2017-02-13
Test

dberzano:


dberzano:
@jgrosseo The CERN username is added after the GitHub user (which is required as we use it to "tag" people), but only in case they differ.

In the example above, my username is the same on GH and at CERN, but Giulio's isn't, and `eulisse` clarifies who `ktf` is :-)

For the moment, the CERN username (and not the full name) is the only information we store.

What do you think?

dberzano:
@jgrosseo I have updated it in order to make the CERN username clickable. Note that searching the CERN Phonebook by username is available only to CERN logged in users, so a wanderer clicking on names on this PR will not be able to access CERN user information.

Do you think we need the link?

dberzano:
+test

dberzano:
(I have deliberately closed this pr.)

PR #50, opened 2017-02-15
Add AliDataFile interface

dberzano:
Also fixes a TPC installation problem (#49)


dberzano:
+1

qgp:
We still need better checks on the string operations, e.g. with version "0-1" you get:
root [2] AliDataFile::Open("test", "")
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 4) > this->size() (which is 3)

qgp:
I am doing some changes now, so maybe let me know before you also start doing something ..

PR #57, opened 2017-02-20
failing commit

jgrosseo:


jgrosseo:
@ktf Nice the log works now! 
Btw. following our chat yesterday, this is an example where one does not see the error in the lines which are posted.

ktf:
@jgrosseo yes. I understand the issue, actually, it's due to the parallel compilation. Now that we have the full log I can make sure the short one has `error:` in it (with some 5 lines context). I was resisting to do it before because I wanted to have the actual output available.

PR #59, opened 2017-02-23
Fix case where no DPMJET is found

dberzano:
DPMJET is disabled if neither external nor builtin DPMJET are found


dberzano:
+1

dberzano:
+1

dberzano:
@qgp Do you agree with the patch? If so, and if it compiles, we can add it to upstream AliRoot.

qgp:
@dberzano: thanks for the fix and, yes, I agree

PR #69, opened 2017-03-17
Update AliBaseLoader.h

sawenzel:
Dummy change to test PR mechanism


sawenzel:
@dberzano, @ktf: I believe that I should not need to be approved for PR testing.

PR #71, opened 2017-03-23
Add .settings in .gitignore

aiola:
It would be good to add .settings in .gitignore. Eclipse users will find it beneficial since it creates this directory automatically.


dberzano:
+1, this is a good idea.

PR #77, opened 2017-03-25
Fix for FMD SDigits->Digits for embedded MC + possibility to increase errors for ITS

shahor02:


ktf:
+test

shahor02:
Hi @cholmcc, apart from trivial change to increment the labels of sdigits from extra inputs (e.g. background event in case of embedding), there was problem in fetching these sdigits: ``fFMD->SetTreeAddress();`` call in AliFMDDigitizer::Digitize always attached tree from the loader attached to AliFMD::fLoader, instead of taking the tree from stream it should process, so the 1st tree was always used. I fixed this by attaching sdigits array to the proper tree branch supplied to SumContributions.

dberzano:
+test

pzhristov:
+1

cholmcc:
+1

However, I think it is messy to let ``AliFMDDigitizer`` set the branch address directly.  It should really be ``AliFMD`` that does this, since that object is the one that manages the digits, etc.

shahor02:
Hi Christian,
Peter has already approved this PR, thanks anyway. You are welcome to reimplement it in a more consistent way, but that into account that the reason why it was not working before (instead of sdigits of extra stream those of the 1st stream were reused) is that AliFMD was handling everything via AliLoader of the current simulation (which indeed knows only the 1st treeS and *must* stay there to handle the treeD, since different treeS are merger to this 1st treeD). Actually, in the original code the digitizer was explicitly receiving the extra treeS and iterating over its entries, if you attach treeS via AliFMD, then I guess the treeS should be also queried from AliFMD.

Cheers,
 Ruben

PR #94, opened 2017-04-01
All tracks will inherit from abstract AliVTrack::GetSign()

shahor02:


pzhristov:
+1

ktf:
This creates some issues in AliPhysics now. I am reverting it while we understand how to proceed.

ktf:
Ok. I've understood why it did not catch the error, at least. For AliRoot PR we currently check AliRoot only. I am changing it to check both AliRoot and AliPhysics.

shahor02:
Hi @ktf, the problem is that this PR is coupled with https://github.com/alisw/AliPhysics/pull/314. My fault that I have not done 1st PR in aliphysics, wait for a couple of weeks until it is merged by some of 20 people authorized to do this, and only then request this PR#95 ... Could you please 1st merge the aliphysics PR then this one?

ktf:
Indeed, we need to make sure you and people like you have a fast track. Not sure why this is not the case already.

dberzano:
Hi @shahor02, you are already an admin with "override" permissions. It means... that you can `+1` every PR even if you are not mentioned.

@pzhristov there is a manually defined `hlt` group in our configuration that we have inherited as it was from Gitolite. The group clearly mentions persons who should not be there indeed. We have to associate it to an egroup, and ask the HLT people to maintain it.

PR #97, opened 2017-04-02
use same settable (Def:300um) threshold for dist. fluctuations

shahor02:
Same threshold for the excess of run-specific distortion dispersion wrt reference one will be used to interpret
it as distortion fluctuation in simulation and reconstruction. Default 300um, can be modified via static setter SetDistDispThreshold.
Used value is reported in the log by UpdateTimeDependentCache


ktf:
+1

PR #100, opened 2017-04-04
Update of AliGenMUONLMR class

bteyssie:
Update of the function Generate and Decay2Body for correction of the kinematics of rho meson

### DIFF
--- a/EVGEN/AliGenMUONLMR.cxx
+++ b/EVGEN/AliGenMUONLMR.cxx
@@ -514,6 +514,7 @@ void AliGenMUONLMR::Generate() {
 	  y   = fY[ipart]->GetRandom(); 
 	  phi = gRandom->Rndm() * 2 * TMath::Pi(); 
 	  mass = pdg->GetParticle(fPDG[ipart])->Mass(); 
+	  if (fPDG[ipart] == 113) mass = fRhoLineShape->GetRandom(); 
 	  px  = pt * TMath::Cos(phi); 
 	  py  = pt * TMath::Sin(phi); 
 	  mt  = TMath::Sqrt(pt * pt + mass * mass);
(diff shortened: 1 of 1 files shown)


rbertens:
+test

rbertens:
+1

@bteyssie could you, in future commits (especially related to 'standard classes' suchs as AliGen*) give a comment which describes the goal of the commit a bit more explicitly? E.g. what does the update provide, what correction? That makes it a bit easier to see how a code change will affect the workflow. 
Thanks in advance!

bteyssie:
Thanks @rbertens, next time I will try to be more precise in the comment associated to the commit. 
Boris

bteyssie:
Hi @rbertens, do you have any idea when the next aliroot version will be released, with the modification of the AliGenMUONLMR class? 
Cheers, Boris

rbertens:
Hi @bteyssie ,
Sorry for the delay in replying. As far as I know, new versions should be proposed in the offline meeting (Mondays at 16 CET) by your PWG convener (the conveners are usually present). If you just want to run, having a new tag suffices, if it needs to be in production code, then I guess one should really requets a new version - this is something to discuss with your convener.

bteyssie:
Hi @rbertens , what I need is an updated AliRoot on cvmfs, so I suppose it is sufficient to have a new tag, is it correct? I this case, which is the expected timeline for the release of the new tag? Do I need to have the agreement of the WG conveners for a new tag?

rbertens:
Hi @bteyssie , as far as i know going through the conveners who propose it at the offline meeting (computing board) is the way to go. Aliroot doesn't have a 'daily' system, so the tagging and deployment over cvmfs has to be initiated by hand and validated. I'm not so sure about the timeline, but perhaps your conveners know more.

PR #104, opened 2017-04-05
Additional fixes for MC-to-MC embedding

shahor02:
1) AliMCEvent::GetParticleAndTR was not aware of subsidiary events
2) EMCAL was applying wrong offset for bg.gigit flagging (10M offset was multiplied by another 10M)


shahor02:
+1

gconesab:
+1

 approuve, although I guess there was a reason for this large factor ... I think it was not me how introduced it.

shahor02:
@gconesab : the offset provided by fDigIput is already multiplied by 10M, no need to do this once more.

PR #106, opened 2017-04-06
using std:...

pzhristov:
As usually such issues are detected by Root6 build, so we have to include it in the validation of pull requests.


pzhristov:
+1

PR #107, opened 2017-04-06
Avoid usage of acos in calculation of Armenteros variables

fprino:
This is a modification to prevent a crash that was spotted in rare case when running AliAnalysisTaskPIDqa in the QA train.
This crash occurs in the calculation of an arc-cosine inside the method AliESDv0KineCuts::Armenteros: in rare cases the argument of the acos function goes out of the [-1,1] range, due to numerical fluctuations. 
The modified code computes the relevant variables without using acos.


shahor02:
+1

PR #108, opened 2017-04-06
CMakeLists.txt: Check for existing -std=c++11 / c++14 flag before add…

aaniin:
…ing it

Hello @ktf @dberzano ,

I realized that currently we compile AliRoot with two times the -std=c++11 flag. When I add output to the main CMakeLists.txt in the AliRoot source like

```
  # Enable C++11 by default if found in ROOT
  if(ROOT_HAS_CXX11 AND NOT DISABLE_CXX11)
    message(STATUS "Enabling C++11")
    message(STATUS "CMAKE_CXX_FLAGS before: ${CMAKE_CXX_FLAGS}")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    message(STATUS "CMAKE_CXX_FLAGS after: ${CMAKE_CXX_FLAGS}")
  endif()

```
One can see in the cmake stage that the -std=c++11 flag is already there in advance and is added a second time here. (I don't know where the first -std=c++11 comes from.)

> DEBUG:AliPhysics:0: -- ROOT was built with C++11 support
> DEBUG:AliPhysics:0: -- ROOT was not built with C++14 support
> DEBUG:AliPhysics:0: -- Enabling C++11
> DEBUG:AliPhysics:0: -- CMAKE_CXX_FLAGS before: -fPIC -g -O2 -std=c++11 
> DEBUG:AliPhysics:0: -- CMAKE_CXX_FLAGS after: -fPIC -g -O2 -std=c++11  -std=c++11

This PR checks whether the -std=c++11 is already in the flags before adding it. Same for C++14.

Cheers,
Hans

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,14 +256,24 @@ else()
 
   # Enable C++11 by default if found in ROOT
   if(ROOT_HAS_CXX11 AND NOT DISABLE_CXX11)
-    message(STATUS "Enabling C++11")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    # Check whether C++11 flag is already in
+    if (${CMAKE_CXX_FLAGS} MATCHES "-std=c\\+\\+11") 
+      message(STATUS "C++11 already enabled")
+    else()
+      message(STATUS "Enabling C++11")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    endif()
   endif()
 
   # Enable C++14 by default if found in ROOT
   if(ROOT_HAS_CXX14 AND NOT DISABLE_CXX14)
-    message(STATUS "Enabling C++14")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+    # Check whether C++14 flag is already in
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -256,14 +256,24 @@ else()
 
   # Enable C++11 by default if found in ROOT
   if(ROOT_HAS_CXX11 AND NOT DISABLE_CXX11)
-    message(STATUS "Enabling C++11")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    # Check whether C++11 flag is already in
+    if (${CMAKE_CXX_FLAGS} MATCHES "-std=c\\+\\+11") 
+      message(STATUS "C++11 already enabled")
```
dberzano: Hi, can we drop the messages saying "already enabled" and make the code more compact? The following would do:

```
message(STATUS "Enabling C++14")
if (NOT ${CMAKE_CXX_FLAGS} MATCHES "-std=c\\+\\+14") 
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
endif()
```

Eventually we are not interested in seeing whether the flag is duplicated or not, we just need to know whether AliRoot understood that we want C++11/14 enabled.

aaniin:
Hello @dberzano 

I changed as requested.

Cheers,
Hans

dberzano:
+1 (and thank you very much, I just did the same in other places)

PR #111, opened 2017-04-09
Fixes needed for proper analysis of MC-to-MC embedded simulation

shahor02:
1) Make sure the labels of mothers/daughters correctly remapped for embedded event
2) Make sure all sub-events are preread in BeginEvent, to build correct offsets table
3) Relable ESD MC lables in case of embedded MC


shahor02:
+1

PR #120, opened 2017-04-11
Adding module structure for EMCAL documentation

mfasDa:
Classes documented already are put into submodules
according to libraries:
- EMCALbase
- EMCALraw
- EMCALrec
- EMCALsim
- EMCALTriggerBase
- EMCALUtils
A top-level module EMCAL groups all submodules. Classes
with fresh documentation will have to be added to the
submodules with the ingroup command.


dberzano:
+1

PR #129, opened 2017-04-12
fix for crash in reconstruction + cosmetic changes

hcab14:


dberzano:
+1

By going through the code I am persuaded that this fixes the problem. As @shahor02 said, storing `TFormula`s in `TClonesArray`s causes a different constructor from being called, which does not invoke `TF1::Optimize()`. The fix here invokes `TF1::Optimize()` manually, before the function is `Eval()`d.

Thanks!

PR #138, opened 2017-04-16
No method Optimize in Root6

pzhristov:
The recent fix breaks the compilation with Root6, so I switched off the method Optimize. It is not available in Root6.


pzhristov:
+1

PR #139, opened 2017-04-16
using -Werr -Wall for AD code

hcab14:
- for rec and sim: TClonesArray::Clear -> TClonesArray::Delete (TF1::Clear is not implemented)
- calling TF1::Optimize() only for ROOT5
- numerous fixes for wrong code formatting
- AliDebug,Info,Fatal,Log(Form(...)) -> AliDebug,Info,Fatal,LogF(...)
- enforcing warning=error for all AD code (tested on OSX/clang, to be verified for gcc4.9)

### DIFF
--- a/AD/ADbase/AliADCalibData.cxx
+++ b/AD/ADbase/AliADCalibData.cxx
@@ -35,37 +35,37 @@
 ClassImp(AliADCalibData);
 
 //________________________________________________________________
-AliADCalibData::AliADCalibData():
-  fLightYields(NULL),
-  fPMGainsA(NULL),
-  fPMGainsB(NULL),
-  fThrCalibA(NULL),
-  fThrCalibB(NULL),
-  fBBAThreshold(0),
-  fBBCThreshold(0) ,  
-  fBGAThreshold(0) ,  
-  fBGCThreshold(0) ,  
-  fBBAForBGThreshold(0) ,  
-  fBBCForBGThreshold(0) ,   
-  fMultADAThrLow(0) ,  
-  fMultADAThrHigh(0) , 
-  fMultADCThrLow(0) ,  
-  fMultADCThrHigh(0)
+AliADCalibData::AliADCalibData()
+  : fBBAThreshold(0)
+  , fBBCThreshold(0)
+  , fBGAThreshold(0)
+  , fBGCThreshold(0)
+  , fBBAForBGThreshold(0)
+  , fBBCForBGThreshold(0)
+  , fMultADAThrLow(0)
+  , fMultADAThrHigh(0)
+  , fMultADCThrLow(0)
+  , fMultADCThrHigh(0)
+  , fLightYields(NULL)
+  , fPMGainsA(NULL)
+  , fPMGainsB(NULL)
+  , fThrCalibA(NULL)
+  , fThrCalibB(NULL)
 {
   // default constructor
-  
+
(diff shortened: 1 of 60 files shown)


pzhristov:
+test

hcab14:
Hi Peter, all,

the compile errors are due to EMCAL and not due to the commit 6a3c3e4, judging from the alibuild log above.

The full log is huge and contains more than 450 aliroot compilations with only ~50% of the log file downloaded.

Cheers
Christoph

pzhristov:
Interesting that the EMCAL test was OK and it compiled also on my Ubuntu machine with Root6.

pzhristov:
+test

ktf:
I am checking.

hcab14:
Hi @ktf,

I forgot to add two files, see bb0a9ec.

Cheers
Christoph

ktf:
The actual error with this PR is above. There is an issue in the error reporting which is making old errors appear again. I am looking into it.

ktf:
+test

ktf:
+test

ktf:
+test

I also found the issue with the huge log file. Will fix tomorrow.

ktf:
+1

mpuccio:
Hi! In my opinion -Werr is a bit error prone for us: the compatibility with clang and gcc 4.9.3 will not ensure the compatibility with all the platforms supported by aliroot (e.g. CentOS7 ships gcc 4.8.5 by default and  currently alibuild/alidist do not require to install gcc 4.9.3 in this case). Moreover the compilation with MacOSX is not tested by the pr checker, thus any future commit in the AD folder might break the compilation for many people silently for compilers/systems supported but not tested (see Lucia's email on the analysis-taskforce mailing list).

shahor02:
Hi @hcab14 

Compilation with gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) fails with:

/home/shahoian/alice/sw/SOURCES/AliRoot/0/0/AD/ADsim/AliAD.cxx: In member function 'virtual void AliAD::Digits2Raw()':
/home/shahoian/alice/sw/SOURCES/AliRoot/0/0/AD/ADsim/AliAD.cxx:405:6: error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]
 void AliAD::Digits2Raw()
      ^
/home/shahoian/alice/sw/SOURCES/AliRoot/0/0/AD/ADbase/AliADRawStream.cxx:99:8: error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]
  Bool_t AliADRawStream::Next()
        ^
Both errors point on the method header, w/o mentioning which may lead to oveflow, looks like result of some deep optimization. Shall we use -Wnerror=strict-overflow ?

hcab14:
Hi @shahor02,

> Both errors point on the method header, w/o mentioning which may lead to oveflow, looks like result > of some deep optimization. Shall we use -Wnerror=strict-overflow ?

Yes. This is reasonable. Indeed this warning occurs with gcc version 5.4.0 (Homebrew GCC 5.4.0_1)  only with option -O2 (not with -O1, -O0 or without optimization).

The pull request is [alisw/AliRoot#149](https://github.com/alisw/AliRoot/pull/149)

shahor02:
Thanks @hcab14, now it compiles w/o problems.

PR #152, opened 2017-04-20
PWGPP-292- Check pointers before deleting

miranov25:
Adding check of the existence of pointer

###  Description:

We are running ITS/TPC/TRD reconstruction benchmark for the LHC15l period with current master  (one day old) (git describe v5-07-20-2281-g023c8b0)
* Out of 370 jobs, 290 jobs failed (see benchmark stacktrace histogram and csv/tree file attached).
* Out of them 220 jobs failed in the AliADReconstructor.
* Not sure if the problem is because of missing check or some memory overweite - but to be sure I added pointer checks 

```
    if (fCorrectForSaturation ) {
    f_Int0->Delete(); // TF1 does not implement TObject::Clear()
    f_Int1->Delete();
      }
```


mbroz84:
+1

PR #154, opened 2017-04-21
Aux.methods for study of dipole field segmentation added

shahor02:


shahor02:
Hi @ktf, @dberzano , the check returned an error but the link to the log is broken. Any idea how to see what is the problem?

ktf:
Most likely a transient problem with one of the build machines which got automatically recovered. I will investigate why we did not get any report.

PR #156, opened 2017-04-22
Fix: TClonesArray must know the class it handles

shahor02:
The ADReconstruction was crashing because the TClonesArray was created w/o specifying the class
it is going to handle, delegating this to tree->GetEntry with branch address set to the array.
But for events when the tree was not read the class was not set and TClonesArray::Delete() was trying to
dereference TClonesArray::fClass w/o checking if it is defined.


shahor02:
+1

PR #163, opened 2017-04-26
AliGenReaderLHE correction

hcab14:
advance to next event record corrected


rbertens:
+1 , and thanks for opening the JIRA ticket, @hcab14

PR #166, opened 2017-04-28
more memory leaks fixed (intermediate)

hcab14:
This is just one more iteration. There are still remaining issues in the code


mbroz84:
+1

PR #167, opened 2017-04-28
added missing dependency XMLIO

hcab14:
for some reason this was not necessary 
- on OSX: ROOT 5.34/30 (heads/v5-34-00-patches@v5-34-28-57-gec27989, Mar 23
2017, 14:28:00 on macosx64), and
- on lxplus7: ROOT v5-34-30-alice7-2
and only showed up when @ekryshen tested AliGenReaderLHE on CentOS7


dberzano:
+1

PR #168, opened 2017-04-30
FORTRAN->C counting convention

hcab14:
see [ALIROOT-7239](https://alice.its.cern.ch/jira/browse/ALIROOT-7239)


rbertens:
+1 @hcab14 is this solution also safe for non-fortran style input ?

hcab14:
Hi @rbertens,

this solution is made for SuperCHIC output. The Les Houches format is specified in terms of FORTRAN common blocks and all information about the LHE format I can find uses FORTRAN numbering.

Anyway I made it user settable, see [AliRoot/pull/169](https://github.com/alisw/AliRoot/pull/169)

It does not cost anything now but may be useful later on.

Cheers
Christoph

PR #169, opened 2017-04-30
made FORTRAN->C counting convention user settable

hcab14:
follow-up to [AliRoot/pull/168](https://github.com/alisw/AliRoot/pull/168)


rbertens:
+1 thanks @hcab14 for the explanation at    #168  .

PR #175, opened 2017-05-04
AD: struct items init compatible with online GCC

dberzano:


dberzano:
+1 (note that this is required to make AliRoot work for legacy SLC6 installations using system's GCC, there was nothing wrong with the code)

PR #179, opened 2017-05-08
Include common methods to AliAODMCParticle and AliMCParticle in AliVParticle for common access of ESD/AOD MC information

gconesab:
I have added to AliVParticle the common methods defined in AliMCParticle and AliAODMCParticle, plus a few more. In doing so would be able to get access the MC information from an analysis task in the same way for ESDs and AODs just doing

AliVParticle* particle = MCEvent()->GetTrack(label)
and replacing AliMCParticle and AliAODMCParticle in the analysis tasks.

Few minor remarks:

* GetStatus() -> There is a different definition in AliVTrack and Ali*Particle, in one it is an Int and in other a long. I have added MCStatusCode() for the MC information

* GetDaughter() -> There is already a method called like that in AliAODRecoDecay returning non int objects, so I renamed if for the recovery of the MC labels as GetDaughterLabel

* T() -> In AliAODTrack there is a template of something with this name so I introduced Tv() like it is done for the coordinates (Xv(), Yv(), Zv()) which returns exactly the same thing as T()

* Momentum(): I use this method from TParticle 

* Particle(): It is only usable for AliMCParticle, but I think it is an easy way to distinguish in analysis code between Stack and AOD information if the pointer is null

### DIFF
--- a/STEER/AOD/AliAODMCParticle.h
+++ b/STEER/AOD/AliAODMCParticle.h
@@ -39,7 +39,9 @@ class AliAODMCParticle: public AliVParticle {
     virtual Double_t Pt()        const;
     virtual Double_t P()         const;
     virtual Bool_t   PxPyPz(Double_t p[3]) const;
-    
+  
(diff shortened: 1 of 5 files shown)


shahor02:
+1

PR #185, opened 2017-05-13
PWGPP-312 - Bug fix: Memory leak

miranov25:
###  PWGPP-312, ATO-373 - Add: Memory diagnostic in AliTreePlayer::MakeHistograms

* Use AliSysInfo (syswatch.log) to indicate memory usage
  * stamp after each histogram creation
  * stamp after each chunk processing
* Print Memory usage in stdout
  * after each histogram creation and total size


shahor02:
+1

PR #197, opened 2017-05-20
possibility to modify seed initial cov.matrix

shahor02:
Matrix used in the very beginning of the Kalman fit can set via pointer
obtained by static AliTPCtrack::GetMatLarge. The last, 16th element of the 
array is optional additive term to q/pt error fC[14]


shahor02:
+1

PR #195, opened 2017-05-19
Fix build on SLC6 for DAQ

dberzano:

PR #198, opened 2017-05-21
Fix once again SLC6 builds

dberzano:


dberzano:
+1

@miranov25 just to be crystal clear:

* Please read more carefully my comment in #195 I say, quote "For the record: I am adding AliRoot test builds on native SLC6 as this is still required by DAQ."
* This means that that PR was a test for the test system, but it was formally correct: code in AliRoot, as per policy discussed countless times, HAS to compile with the native SLC6 compiler. So I am restoring what I have done in #195, and you have reverted in #197.
* We will have at some point automatic tests to catch this, we have had some problems integrating them, so no blame here - it's us who have to automatically catch such problems :-)

Cheers

d.

STAT/AliDrawStyle.cxx
```diff
@@ -61,11 +61,11 @@
 //
 std::map<TString, TString>  AliDrawStyle::fLatexAlice;
 std::map<TString, TStyle*>  AliDrawStyle::fStyleAlice;
-std::map<TString, std::vector<int>> AliDrawStyle::fMarkerStyles;
-std::map<TString, std::vector<int>> AliDrawStyle::fMarkerColors;
-std::map<TString, std::vector<float>> AliDrawStyle::fMarkerSize;
-std::map<TString, std::vector<int>> AliDrawStyle::fFillColors;
-std::map<TString, std::vector<float>> AliDrawStyle::fLineWidth;
+std::map<TString, std::vector<int> > AliDrawStyle::fMarkerStyles;  // PLEASE LEAVE THE UNAESTHETIC SPACE
```
ktf: How come AliRoot builds with C++11?

miranov25: Hallo.

>We will have at some point automatic tests to catch this, we have had some problems integrating them, so no blame here - it's us who have to automatically catch such problems :-)

Today I will continue with modification of class. I would like to ask you how to check my code before pull request.  I expect (maybe I'm wrong) there are not so many cases which we do not catch with default  compiler but are problems for other compilers (ubuntu/SLC/Mac....)
 
Can you provide list of the sed regular expression to indicate, eventually fix errors and aesthetic changes? 
E.g before my commit I removed all space at the end.
 ```
sed 's/ *$//' 
```


Marian.

PR #208, opened 2017-05-25
Add missing using namespace std

mfasDa:
Needed for compilation under ROOT6.


wiechula:
Isn't it better to do explicit 
using std::cout;
using std::endl;
using std::multiset;
instead of including the full namespace?

miranov25:
+1
for future  I also prefer to use std:: instead using namespace

PR #214, opened 2017-05-30
fix channel masking, increase verbosoty

wiechula:
In the implemented logic we relied on one point at SOR and one at EOR
At some point the preprocessor was changed in a way that deletes points
outside a certain dead band, which also most of the time deletes the
point at EOR.
The logic needed to be update to be able to deal with this

### DIFF
--- a/TPC/TPCbase/AliTPCcalibDB.cxx
+++ b/TPC/TPCbase/AliTPCcalibDB.cxx
@@ -993,7 +993,7 @@ Int_t AliTPCcalibDB::InitDeadMap()
     }
 
     if (numberOfDeactivatedChannels>0) {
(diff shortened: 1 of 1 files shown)


miranov25:
It is good that the logic was fixed, so it  can be used.
However we should inform responsible that this behavior is not expected, moreover it is not desired.
We should get back our bonus points before and after the run.

wiechula:
Hi Marian, the point at SOR we have, the one at EOR is thrown away by the dead band.
Perhaps Haavard can change the logic in the preprocessor in such a way that the last point in the graph will always be kept. This should be trivial.

miranov25:
Hallo.
Yes. Lets keep all "bonus values"

PR #216, opened 2017-05-31
Fix DPMJET CMake warning

dberzano:


dberzano:
@loizides thanks for spotting this

PR #220, opened 2017-06-01
T0 dev

AllaMaevskaya:
CFD mean position vs timestamp added to CPass0 treatment

T0A, T0C, T0AC and resolution as a function of event multiplicity (N contributers to vertex)


chiarazampolli:
Ciao Alla,

Is this needed for CPass? If so, please, open a JIRA porting request.

Cheers,

Chiara

PR #221, opened 2017-06-01
Hiing econv

loizides:
Patch makes possible to disable adhoc energy conservation
and allows the number of misses to be set in hijhrd.f
The settings are propagated via IHPR2(31) and (30) using
the function AliGenHijing::SetEconv(Bool_t ec, Int_t nmiss=50).

The patch also removes annoying print outs to fort.16 and fort.8


loizides:
Of course if applied then I would also want this to be included in the next aliroot tag.

pzhristov:
+test

pzhristov:
+1

PR #232, opened 2017-06-07
Enable optional CVMFS OCDB usage

dberzano:
This feature is required for accessing the OCDB from CVMFS expecially during the
MC release validation. OCDB is automatically accessed from there, if:

* `AliCDBManager::SetDefaultStorage("raw://")`
* and environment variable `OCDB_PATH` is set to a proper CVMFS path, e.g.
  /cvmfs/alice-ocdb.cern.ch

Run mapping to the correct folder is performed the same way as on AliEn, by
using the same XML mapping file. Mapping is not performed in case
`SetDefaultStorage` is set to a specific directory.

Those changes are backwards-compatible: the previous behaviour is unchanged in
case the `OCDB_PATH` variable is not exported. CVMFS OCDB must be enabled
explicitly.

### DIFF
--- a/STEER/CDB/AliCDBLocal.cxx
+++ b/STEER/CDB/AliCDBLocal.cxx
@@ -826,24 +826,12 @@ void AliCDBLocal::QueryValidFiles() {
 // Fills list fValidFileIds with AliCDBId objects extracted from CDB files
 // present in the local storage.
 // If fVersion was not set, fValidFileIds is filled with highest versions.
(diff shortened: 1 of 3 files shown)


dberzano:
Credits to @mconcas. @preghenella: this goes together with alisw/AliDPG#31.

preghenella:
Thanks, @mconcas and @dberzano, looking forward to using this

PR #234, opened 2017-06-07
Geometry updates for AD detector

ecalvovi:
This updates geometry of several structures surrounding ADA and ADC detector. Most importantly the dipole compensator magnet on A-Side.

PR #238, opened 2017-06-08
Geometry updates for AD

sawenzel:
* Corrected Dipole geometry for ADA (2017)
 * Ovalized Beam Pipe (RB242) inside Dipole
 * A-Side: Shielding for PM's, I-Beam and other beam-pipe supports, OldAD, BLM's, support for Warm Module
 * C-Side: Grid of U-Profiles near the wall that separates CAVERN from LHC tunnel

supercedes #234.

### DIFF
--- a/AD/ADsim/AliAD.cxx
+++ b/AD/ADsim/AliAD.cxx
@@ -65,7 +65,7 @@
 #include "AliADCalibData.h"
 #include "AliADConst.h"
 
-ClassImp(AliAD);
+ClassImp(AliAD)
 //__________________________________________________________________
 AliAD::AliAD()
   : AliDetector()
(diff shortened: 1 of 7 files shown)


mbroz84:
+1

chiarazampolli:
Hello,
In the other porting request, it was said that this is needed for simulation. Does it mean that previous MC are affected by some sort of bug?
Cheers,
Chiara

ecalvovi:
Hi CHiara, 

The forward geometry description in AliRoot was incomplete/incorrect. This commit fixes that and cover a z range of aprox [5; 17] meters on the A-Side and aprox [-18;-19] meters on the C-Side. So this does only affect AD and maybe ZEM, other detectors are not affected. The biggest discrepancy I would say is in the geometry of the compensator magnet in A-Side at 10.75 meters from IP. 

Preliminary local simulations show that the effect is small even for AD. This maybe be important however for diffractive studies.

chiarazampolli:
Ok, thanks!
Chiara

PR #241, opened 2017-06-08
Do not fill with 0s freshly create TObjArray

shahor02:
This is to silence annoying ``E-TObjArray::operator[]: index 16 out of bounds...`` errors, the bug itself was malign


shahor02:
+1

PR #242, opened 2017-06-08
Fix of non-initialized var. + memory leak

shahor02:
The non-initialized variable was incremented in loop and used as the size of TGraph, potentially the reason of v5-09-09 failure: https://alice.its.cern.ch/jira/browse/ALIROOT-7321?focusedCommentId=197457&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-197457


shahor02:
+1

PR #244, opened 2017-06-09
EVE: char* -> const char* fix in header

aaniin:
Fixes char* -> const char* for string literal in header which otherwise gives plenty of warning messages.  Can you comment @jniedzie ?


jniedzie:
+1

PR #245, opened 2017-06-09
update AliGenPythiaPlus with same way to select calorimeter acceptanc…

gconesab:
…e and particles as in AliGenPythia.
This will allow to use Pythia8 in the same way as Pythia6 to trigger the events on high pT decay photons like in past MC productions.

### DIFF
--- a/PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
+++ b/PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
@@ -35,9 +35,7 @@
 #include "AliLog.h"
 #include "PyquenCommon.h"
 
-/// \cond CLASSIMP
-ClassImp(AliGenPythiaPlus);
-/// \endcond
+ClassImp(AliGenPythiaPlus)
 
 AliGenPythiaPlus::AliGenPythiaPlus():
     AliGenMC(),
@@ -1346,10 +1344,7 @@ void AliGenPythiaPlus::GetSubEventTime()
 ///
 Bool_t AliGenPythiaPlus::IsInBarrel(Float_t eta) const
 {
-  if( eta < fTriggerEta  ) 
-  return kTRUE;
-  else 
-  return kFALSE;
+  return ( eta < fTriggerEta ) ;
 }
 
 ///
@@ -1380,7 +1375,7 @@ Bool_t AliGenPythiaPlus::IsInPHOS(Float_t phi, Float_t eta, Int_t iparticle)
 {
   if(phi > fPHOSMinPhi  && phi < fPHOSMaxPhi && 
      eta < fPHOSEta  ) 
-  return kTRUE;
+    return kTRUE;
   else 
   {
     if( fCheckPHOSeta && eta < fPHOSEta) fPHOSRotateCandidate = iparticle;
--- a/PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
+++ b/PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
@@ -13,17 +13,6 @@
  * provided "as is" without express or implied warranty.                  *
  **************************************************************************/
 
-/* $Id$ */
-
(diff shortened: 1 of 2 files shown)


gconesab:
Hello, can someone at least do +test?
Here I just copy pasted and adapted what is present already in AliGenPhythia (Pythia6) so that the same tricks can be applied on Pythia8.

ktf:
+test

amorsch:
The changes look reasonable to me.
Andreas

gconesab:
Thanks Andreas, can someone approve +1?

sawenzel:
I have a general remark: It seems to me that many changes are just reformatting of the code which makes it a bit hard to identify the actual things done.

PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
```diff
@@ -43,10 +32,12 @@
...
 #include "AliRunLoader.h"
 #include "AliMC.h"
+#include "AliLog.h"
 #include "PyquenCommon.h"
 
-ClassImp(AliGenPythiaPlus)
-
+/// \cond CLASSIMP
+ClassImp(AliGenPythiaPlus);
+/// \endcond
```
sawenzel: Is this change necessary? Do we need the `;`?

PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
```diff
@@ -1381,81 +1339,365 @@ void AliGenPythiaPlus::GetSubEventTime()
...
+///
+/// \param eta        particle pseudorapidity,  etamin=-etamax
+///
+Bool_t AliGenPythiaPlus::IsInBarrel(Float_t eta) const
+{
+  if( eta < fTriggerEta  ) 
+  return kTRUE;
+  else 
+  return kFALSE;
+}
```
sawenzel: consider using `return (eta < fTriggerEta);`

pzhristov:
I have couple of comments:
- in two places I see semicolon after ClassDef or ClassImp. Please remove them, these are Root macros and semicolon would generate warnings on some platforms
- The code below
+Bool_t AliGenPythiaPlus::IsInBarrel(Float_t eta) const
 +{
 +  if( eta < fTriggerEta  ) 
 +  return kTRUE;
 +  else 
 +  return kFALSE;
 +}

can be replaced by a single line 

return (eta < fTriggerEta);

PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
gconesab: This is needed for doxygen, I can remove it.

There are only a couple of reformating, most of the changes are replacing one old piece of code by another.

PYTHIA6/AliPythia6/AliGenPythiaPlus.cxx
gconesab: It is a copy paste from AliGenPythia, I can change it here and there or only here.

sawenzel: Thanks. So it seems that we have a lot of common functionality between the 2 classes. The proper thing would probably be to factor out common code or dispatch from one to the other.
[This is something we can keep in mind for AliceO2 ... not asking to completely change the classes now.]

gconesab: I thought of doing this, but preferred simplicity and there are subtleties between the 2 generators that cannot be treated similarly (status code only 0 or 1 in pythia8, label index+1 in pythia8). I will do the requested changes

gconesab:
Requested changes done, please approve or reject.

sawenzel:
+test

gconesab:
hello, can someone approve?

pzhristov:
+1

PR #248, opened 2017-06-12
PWGPP-312 - Add support for nested json using root TBufferJSON

miranov25:
###  Support for Elastic bulk export
*  For elastic export \n has to be removed (_bulk export in elastic per line)
```
obuffer = TBufferJSON::ConvertToJSON(*ppobject, isClass[icol]);
if (isElastic) obuffer.ReplaceAll("\n","");
```
* Adding test for the bulk export with nested classes

### Missing/ToDo
* **jq filters for listing and elastic->tree for nested classes not yet implemented** 
(first we provide input documents)
*  Exporting nested classes the size of the document can explode
   * e.g exporting of full QA.TPC tree is not possible because of too many data members to represent TGraphErrors (metadata with axis description, histogram, list of functions...)
   * Light JSON to be parameterized.  Typically array of fY,fX,fEY,fEX should be sufficient. Metadata to be stored separately

### DIFF
--- a/STAT/AliTreePlayer.cxx
+++ b/STAT/AliTreePlayer.cxx
@@ -16,7 +16,7 @@
 
 /*
   Set of functions to extend functionality of the TTreePlayer
-  data source (tree+friend trees)  configured independently in AliExternalInfo
+  - data sources (tree+friend trees)  configured independently in AliExternalInfo
 
   Functionality:
   * function to support metadata and collumns annotation
@@ -50,6 +50,7 @@
 #include "TEntryList.h"
 #include "THn.h"
 #include "TLegend.h"
+#include "TBufferJSON.h"
 #include "AliSysInfo.h"
 
 using namespace std;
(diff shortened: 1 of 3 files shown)


miranov25:
can somebody check the pull request and comment?
As i wrote in previous pull request, Elastic interface which I implemented is  limited current root/C++ restriction of AliRoot software.
Once we will start to use C++11 or higher we can use C++ implementation instead of the (curl+jq+TBufferJSOn) implementation
What is important this interface exist and I provide root->Elastic support:
* tree->elastic DB  working
* elastic->tree  -support for nested json not yet working

I can do development in special branch but in that case I will have problems to get user feedback.
This solution can also slow down my other activities.

pzhristov:
+test

pzhristov:
+1

dberzano:
Hi all, note that `TBufferJSON.h` requires we build ROOT with HTTP support, which currently might, or might not happen (we do not check for it). This commit [breaks compilation of DAs](https://alijenkins.cern.ch/job/DAQ/job/daq-da-rpms/406). For the next AliRoot release we will make sure the recipe contains it.

PR #250, opened 2017-06-13
include charged kaons in AliDecayerPythia long lived decays

laltenka:
in order to let them decay using the Pythia6 decayer (like e.g. K0 long)


dberzano:
+test it looks fine to me, but is there someone who can sign this off?

PR #251, opened 2017-06-14
Updated RemoveGraphDuplicates so that always first and last points are kept

haavardh:
On request from Jens Wiechula, the removal of duplicate points (or rather points falling within a tolerance interval) is changed so that the first and last point are always kept. This just required modifications to the loop control statement (and the initial protection test) -- no other change to the logic.


dberzano:
+test @haavardh is there a related JIRA to that?

haavardh:
No. I can of course make one. The request was discussed at a TPC meeting.

dberzano:
As long as it's signed off by the TPC experts this is fine, can we have either a JIRA or a comment here from one of them? Thanks!

haavardh:
OK. I have asked them to comment here.

dberzano:
+1 thanks @haavardh and @wiechula!

PR #259, opened 2017-06-15
Disable JSON when ROOT HTTP is not available

dberzano:


dberzano:
+test (do not merge yet!)

dberzano:
@pzhristov @miranov25 I had to disable the only class using JSON from ROOT's HTTP when the latter is not available.

Note that by default JSON will *always* be available, except when building DAs, where we do not need it. The measure was necessary as DAs are built statically, and libRHTTP from ROOT does not play well if statically included - plus it considerably increases the DA size for an unneeded feature.

Just bear in mind to use the same "disabling" techniques I have used in case further classes requiring JSON are added in the future :-)

dberzano:
Please hold on and do not approve yet.

dberzano:
+1

pzhristov:
Hi Dario,
Thank you very much, I forgot that the DA are using static library and did not test this case.

Cheers,
Peter

PR #262, opened 2017-06-20
Macros to create and update OADB with L1 phases and time calibration

amatyja:
1. Macro to create OADB root object with L1 phases for EMCal.
2. Macro to update OADB root object with L1 phases for EMCal.
3. Macro to update OADB root object with time calibration for EMCal.


gconesab:
+1

PR #265, opened 2017-06-21
Protection against Short_t overflow when filling correction maps

shahor02:
@fprino : this is to protect correction maps from the sign flip due to the Short_t (used internally to store the map) overflow, which happens in very rare cases of "nonuniformiry" exceeding ~3.2mm (mostly artifact of modules with only fraction of the drift distance having calibration data)


fprino:
Hi Ruben,

thanks! 

Cheers, Francesco

PR #267, opened 2017-06-22
Protect against missing information - such as ESD tracks or AliMultip…

cholmcc:
This patch will protect against segmentation violations if an AliESDtrackCuts object is used on an 
event that for example, does not have tracks or the SPD multiplicity object.


dberzano:
+1

hcab14:
Hi Christian,

I have a minor a comment about cosmetics: in line 3030 and following the indentation is wrong (at least it looks wrong in the web browser) which makes it hard to read this code.

Christoph

cholmcc:
Hi Christoph,

Fixed the indention.  But really? The code is sooo poorly formatted in the first place (lines extending way beyond column 80, overly many spaces, and so on) that it probably didn't really matter :-p 

Christian

cholmcc:
I all,
Perhaps someone need to +1 (or +test) this again after I fixed the indention?

shahor02:
+1

PR #268, opened 2017-06-22
Recognize root_archive*.zip as Zip archives.  This is useful if one h…

cholmcc:
This patch will allow the input handler to recognise ZIP archives that has names like 
root_archive_X.zip (where X may be a sequence number - for example).  This allows users 
 to store local mirrors of data in a single directory.

### DIFF
--- a/STEER/ESD/AliESDInputHandlerRP.cxx
+++ b/STEER/ESD/AliESDInputHandlerRP.cxx
@@ -99,7 +99,8 @@ Bool_t AliESDInputHandlerRP::Init(Option_t* opt)
     {
       TString rppath;
       if (det->TestBit(kReadFromArchiveBIT)) {
-	if (fPathName->EndsWith("root_archive.zip"))
+	// if (fPathName->EndsWith("root_archive.zip"))
+	if (fPathName->Contains("root_archive"))
 	  rppath.Form("%s#%s.RecPoints.root", fPathName->Data(), det->GetName());
 	else
 	  rppath.Form("%sroot_archive.zip#%s.RecPoints.root", fPathName->Data(), det->GetName());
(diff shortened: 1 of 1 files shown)


STEER/ESD/AliESDInputHandlerRP.cxx
```diff
@@ -99,7 +99,8 @@ Bool_t AliESDInputHandlerRP::Init(Option_t* opt)
     {
       TString rppath;
       if (det->TestBit(kReadFromArchiveBIT)) {
-	if (fPathName->EndsWith("root_archive.zip"))
+	// if (fPathName->EndsWith("root_archive.zip"))
+	if (fPathName->Contains("root_archive"))
```
dberzano: Would it work if you checked both if it contains "/root_archive" and ".zip" (with, without a regexp I don't care). Also, can you just delete the commented line?

STEER/ESD/AliESDInputHandlerRP.cxx
```diff
@@ -99,8 +99,9 @@ Bool_t AliESDInputHandlerRP::Init(Option_t* opt)
     {
       TString rppath;
       if (det->TestBit(kReadFromArchiveBIT)) {
-	// if (fPathName->EndsWith("root_archive.zip"))
-	if (fPathName->Contains("root_archive"))
+	if (fPathName->Contains("root_archive", TString::kIgnoreCase) &&
+	    fPathName->Contains(".zip", TString::kIgnoreCase))
```
dberzano: `EndsWith`?

cholmcc: No, EndsWith would exclude other cases, such as 
```
root_archive.zip.001
```
I think the library is agnostic to the _ending_ so it shouldn't really matter. 

Of course, by far the best option would be _not_ to rely on the file name but instead use file magic - but alas ROOT does not provide that service. 

Christian

dberzano: OK you are right, thanks for clarifying.

dberzano:
+1

PR #269, opened 2017-06-22
Include DIM header as <dim/dim.hxx> since the header is always

cholmcc:
Include the DIM header as <dim/dic.hxx> as it is always in a sub-directory named dim.

### DIFF
--- a/HLT/createGRP/lib/AliHLTCreateGRP.cxx
+++ b/HLT/createGRP/lib/AliHLTCreateGRP.cxx
@@ -16,7 +16,7 @@
 //***************************************************************************
 
 #include "AliHLTCreateGRP.h"
-#include <dic.hxx>
+#include <dim/dic.hxx>
 #include <stdio.h>
 #include <ctime>
 #include <iostream>


pzhristov:
Hi @cholmcc,

In CMake we have the following line:

cmake//FindDIM.cmake:19:    find_path(DIM_INCLUDE_DIR NAMES dim.h PATHS ${DIMDIR}/dim NO_DEFAULT_PATH  DOC "Path to DIM header folders.")

so the include path is actually to the subdirectory where we have dim.hxx. I suspect that with your change we would need also to modify the CMake file. Do we really need it knowing that this part of AliRoot has been used in the same way since more than 10 years?

dberzano:
@pzhristov is right. In addition we have (finally, thanks to @Barthelemy) a recipe for DIM so that we know how we build it. See it [here](https://github.com/alisw/alidist/blob/master/dim.sh).

This recipe, out of compatibility with the current (admittedly, weird) way of doing things, installs headers under `${DIMDIR}/dim`, and it is compatible with the CMake line quoted above. The include path as added by CMake already includes the `dim` subdirectory.

Closing this.

PR #271, opened 2017-06-22
Changes to do with primary particle definition as implemented

cholmcc:
in `AliStack::IsPhysicalPrimary`

- The code of `AliStack::IsPhysicalPrimary` was changed to conform with
  the definition adopted by ALICE in a recent public note:

>    A primary particle is a particle with a mean proper lifetime
>    tau larger than 1cm/c, which is either a) produced directly in the
>    interaction, or b) from decays of particles with tau smaller
>    than 1cm/c, restricted to decay chains leading to the interaction.

  See also ALICE-PUBLIC-2017-005 at

     https://cds.cern.ch/record/2270008

  Please note, that the definition is approved by the *Physics Board*

- The old implementation of `AliStack::IsPhysicalPrimary` is still
  available as `AliStack::IsPhysicalPrimaryOld` to allow testing etc.

- Various bugs in the old implementation (now `AliStack::IsPhysicalPrimaryOld`)
  and in `AliStack::IsSecondaryFromWeakDecay` were corrected.  Specifically
  these bugs had to do with extracting the heaviest flavour or a
  particle (to see if a particle was heavy flavour or strange)

  To get the heaviest flavour of a particle with PID pdg the old code
  tried to extract the heaviest flavour quark by doing

    `f = floor(|pdg| / 10^{floor(log10(|pdg|)})`

  Value of f corresponds to the heaviest quark in the particle
  (1: d, 2: u, 3: s, 4: c, 5: b, 6; t) For most particles
  this is fine, e.g.,

  - pi+:   pdg=211  -> f=2
  - K+:    pdg=311  -> f=3
  - K0(S): pdg=310  -> f=3
  - K0(L): pdg=130  -> f=1(!)
  - p      pdg=2212 -> f=2
  - B+:    pdg=521  -> f=5
  - D+:    pdg=411  -> f=4

  while for others it will give the wrong result

  - D(0)*      pdg=10421   -> f=1 (should be 4)
  - K(1)+      pdg=10323   -> f=1 (should be 3)
  - f(0)(500)  pdg=9000221 -> f=9(!) (should be 2)

  and the leptons are really messed up

  - e-   pdg=11 -> f=1
  - mu-  pdg=13 -> f=1
  - tau- pdg=15 -> f=1

  The definition of the PDG code conventions state that each
  PDG code is formatted as

>    n nr nL nq1 nq2 nq3 nJ

  where _nq1_, _nq2_, and _nq3_ encode the quark content.
  For mesons _nq1_=0.  So a better way to extract the heaviest
  flavour is to take the absolute value of the PDG code, modulo
  10,000 and then, depending on whether the value is larger than
  999 (baryons) extract the 4th or 3rd digit

```
    if (|pdg|<100 && |pdg|>6) return 0; // No quark content
    m = |pdg| % 10000; // get nq1 nq2 nq3 nJ
    if (m == K(0)L) return 3; // special for K0L
    f = (m > 999) ? m / 1000 : m/100;
```

  Note, for nuclei the contvention is to encode as

>     10LZZZAAAI

  where _L_ is the number of Lambda's in the nucleus.  So we
  should check if this number is non-zero and in that case
  get 3 (strange), otherwiese 2 (up).  I.e., the full
  pseudo-code is

```
    if (|pdg|>1000000000)
      f = (|pdg|/10000000) % 10) != 0 ? 3 : 2;
    else if (|pdg|<100 && |pdg|>6)
      f = 0; // No quark content
    else
      m = |pdg| % 10000; // get nq1 nq2 nq3 nJ
      if (m == K(0)L) return 3; // special for K0L
      f = (m > 999) ? m / 1000 : m/100;
```
  This algorithm is implemented in the service function
  `HeaviestFlavour` in the scope of *AliStack.cxx*

- Other small efficiency fixes.

### DIFF
--- a/STEER/STEERBase/AliStack.cxx
+++ b/STEER/STEERBase/AliStack.cxx
@@ -954,6 +954,79 @@ void AliStack::ConnectTree(TTree* tree)
     AliWarning("Branch Dir is NOT SET");
 }
 
+// Some utilities used 
+namespace {
+    /** 
+   * Check the heaviest flavour of a particle.  Note, the original
+   * implementation is
+   *
+   *@f$f =  \left\lfloor\frac{c}{10^{\lfloor\log_{10}c\rfloor}}\right\rfloor@f$
+   * 
+   * since it will return 1 for @f$\Upsilon(2\mathrm{S})@f$ (@f$ c=100553@f$). 
+   * 
+   * The coding convention for PDG numbers is 
+   *
+   * @f$ n n_r n_L n_{q1} n_{q2} n_{q3} n_J@f$ 
+   *
+   * which means to guage the quark content we need only the 4
+   * right-most digits.  The most optimal way to extract these is to
+   * format the modulo of the passed number to @f$10\,000@f$, format
+   * that remainder as a string and then take the first character and
+   * convert that to a number by subtracting off the character @c '0'.
+   * 
+   * @param pdg 
+   * 
+   * @return 
+   */
+  Int_t HeaviestFlavour(Int_t pdg)
+  {
+    Int_t apdg = TMath::Abs(pdg);
+    // Special codes for bosons and leptons 
+    if (apdg < 100 && apdg > 6) return 0;
+    // Special codes for nuclei
+    // Ion/nucleus with format 10LZZZAAAI - take the 10L part modulo
+    // 10 to get the L part.  Hyper nucleons with np protons, nn
+    // neutrons, and nl Lambdas have
+    //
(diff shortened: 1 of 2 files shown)


dberzano:
+1
thanks for specifying that this is approved by the PB :-)

amorsch:
The while-loop in AliStack::IsPhysicalPrimary can go back to the colliding particles which are stable. 
In this case the particle is falsely flagged as a secondary. This happens quite frequently. We have to switch back to the old version of the code until this is fixed.

PR #276, opened 2017-06-22
[ALIROOT-7344]: Removing access to a dereferenced nullptr

sawenzel:
* Urgent fix removing a wrong statement
 * Reenables the vmctest/gun to run correctly
 * The original authors should cross-check and eventually provide the
   correct behaviour


mfasDa:
Clearly this is a bug and need to have a closer look by the original author. This fix is a temporary fix. For the current purpose it should not harm, so I will approve it.

mfasDa:
+1

PR #278, opened 2017-06-23
Fix merge from .zip not from AliEn

dberzano:
Merging .zip files from non-AliEn paths requires specifying the inner .root file
with an anchor, in the format `arch.zip#file.root`. On AliEn this is not
necessary as the .zip entries are stored in the catalog as part of the current
directory.

It also tells the TFileMerger to access the .zip files not on AliEn directly, as
opposed to copying them locally first, as `TFile::Cp()` does not support .zip
anchors (silently).


dberzano:
+1

dberzano:
@mconcas @preghenella this is one more thing (the last, I hope, after #232) that we need for the MC validation (I am surprised this problem did not come up earlier).

PR #281, opened 2017-06-23
Updated default cut on Ncontributors in IsPileupFromSPD

fprino:
This commit modifies the default values for the pileup tagging with SPD changing in particular the minimum contributor cut from 3 to 5.
The reason is that the cut at 3 contributors was seen to introduce a biased at high multiplicity due to positive taggings  (introducing a distortion in the multiplicity distribution).

This change was discussed in AOT-event properties meeting (@ddobrigk) and announced via email some weeks ago. Nobody opposed.
We verified that the selection at 5 contributors is already the one used in the physics selection (@ekryshen can confirm) and that this change will not affect the selections applied in the multiplicity task (@ddobrigk).
Note also that in AliAnalysisUtils (the wrapper class for event selection used for many analyses @mfloris) the cut is set by default at 5 contributors.
Add @shahor02 and @chiarazampolli in the discussion.


dberzano:
+1 (discussed in the DPG)

PR #283, opened 2017-06-24
test AOD filtering must use reference OCDB

shahor02:
AOD filtering in test directories should use the OCDB described in
http://alisw.github.io/git-advanced/#use-the-reference-ocdb


shahor02:
+1

PR #286, opened 2017-06-26
Emcal trigger testing

moliver813:
Long term fix for case where Run 2 EMCAL geometry with DCAL is requested, but DCAL STU Config is missing from OCDB.  Clones the EMCAL STU Config and changes the patch size and firmware to valid choices for the DCAL.  Under proper conditions, this scenario should not occur.


mfasDa:
Please do not use Clone. Rather fix the copy constructor of AliEMCALTriggerSTUConfig.

moliver813:
It no longer uses Clone for copying the EMCALSTUDCSConfig

gconesab:
+test

EMCAL/EMCALbase/AliEMCALTriggerSTUDCSConfig.h
```diff
@@ -86,9 +86,6 @@ class AliEMCALTriggerSTUDCSConfig : public TObject
   void    GetSegmentation(TVector2& v1, TVector2& v2, TVector2& v3, TVector2& v4) const;
   
   TClonesArray *GetErrorCountsForTRU(Int_t itru) const;
-  
-protected:
```
gconesab: I think the removal of the protected is the cause of the failure, you should add it at least for the = operator that you have not created, but I am not sure, you could try it

mfasDa: Indeed the issue seems to be connected with a missing implementation of the operator= (which becomes declared now moving it from protected to public). This can be implemented similarly to the copy constructor.

gconesab:
Michael, can you do the fix today?I would like to ask to include this in the next aliroot tag. Maybe, if it takes too much time, no need to implement the =operator but make it protected again, not the copy constructor.

moliver813:
Just made the = operator protected again.  Sorry, I didn't realize making it public would require implementation.

gconesab:
+test

gconesab:
+1

PR #293, opened 2017-06-29
ATO-245 - Bug fix - reset variable py->SetMINT(51,0) 

miranov25:
ATO-245 - Bug fix - reset Pythia common block py->SetMINT(51,0); ￼…
* Add AliSysInfo::AddStamp
* Reset common block (see discussion in ATO-245) after each event
  * py->SetMINT(51,0);
  * Why/What change common using new AliRoot/AliDPG is not clear
* Momenutm conservation (pz)
* Update SetHighWaterMark(nPart) only in case new particle added to stack

To come (after discussion with AliDPG)
* decay products enhancement or branching ratios
* enable pythia configuration  (e.g pi0 decay)

### DIFF
--- a/EVGEN/AliGenPerformance.cxx
+++ b/EVGEN/AliGenPerformance.cxx
@@ -66,6 +66,8 @@
 #include <AliPythia.h>
 #include <TTreeStream.h>
 #include <TChain.h>
+#include "AliSysInfo.h"
+
 
 ClassImp(AliGenPerformance)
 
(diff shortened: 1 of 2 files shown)


miranov25:
Hallo 
 @rbertens   @qgp 
Can you please approve bug fix in performance generator.
I would like to start test production with an hour

miranov25:
+test

miranov25:
Hallo  @amorsch 
could you please approve pull request for AliGenPerformance ?
Looks like Jochen and Redmer are not arround.

pzhristov:
OK, I checked and discussed with Marian on Skype. It looks to me that the generation will have to be changed again after the test production. I am going to approve this pull request.

pzhristov:
+1

EVGEN/AliGenPerformance.cxx
```diff
@@ -88,35 +90,35 @@ AliGenPerformance::AliGenPerformance():
...
-  fVerboseLevel(0)            // verbose level
+        AliGenerator(),
+        fNJets(1),                // mean number of jets per event
+        fF1Momentum(0),           // momentum distribution function
+        fFPhi(0),                 // phi distribution function
+        fFTheta(0),               // theta distribution function
+        fFPosition(0),            // position distribution function
+        fFPdg(0),                 // pdg distribution function
+        fStreamer(0),           // test stream - used for tuning of parameters of generator
+        fVerboseLevel(0)            // verbose level
```
sawenzel: why these changes?

rbertens:
hi @miranov25 , sorry for the slow response, i was teaching

EVGEN/AliGenPerformance.cxx
pzhristov: This is just indentation according to Clion. The real changes are very few. You can see them if you add ?w=1 at the end of the URL

sawenzel: I know that I can see them with ?w=1. But I believe these changes are not necessary and only cause confusion with `git blame` etc. I would like to express again my opinion that we should try to avoid reformatting changes altogether and definitely not mix them with real changes.

miranov25: @sawenzel 
We discussed with Dario re-formatting issue some time ago but we did not conclude.

What I did in this commit:
I recently started to use CLIon for editing and debugging. 
Unfortunately old default indentation of xemacs is not compatible with CLion indentation  - and code was difficult to read. I applied CLion automatic indentation. 

I would like to ask following question:
* Can we define default AliRoot formatting?
  * I would like to apply this formatting in CLIon and xemacs
*   Should we reformat class by class, before actual modifying of the code?

I prefer to conclude as soon as in the next code modification I will have again the same problem.

Marian

sawenzel: There does not seem to be a `.clang-format` file which would be good to have. But I thought the styling is specified somewhere even for AliRoot.

What I am saying is
a) we should avoid user A applying style from his editor and user B changing it back to his setting
b) if formatting is done it makes sense to pack it into a separate commit so that is clear to other developers ... (or at least mention it clearly in the commit message).
c) We should configure our editors not to do formatting by default.

sawenzel:
@dberzano , @ktf: It seems that one the PR tests is not finishing and idling since yesterday (same for other PRs).

miranov25:
do you know why the test was not yet done?
Is there some problem with server?

EVGEN/AliGenPerformance.cxx
miranov25: @sawenzel 
Can we move further discussion to JIRA as it is not directly related to pull request?

Before JIRA task will be created, here are my consideration:
I think we agree we need some default formatting. 
For your point a..c they are not easy applicable.

* Currently I'm both users from your example: user  A (CLIon editor) and user B (XEmacs editor)
  * in my commit I used auto-intent of CLion
* To avoid problems with new code I have to know which formatting to use
  * it is not important what formatting we will use, but it is should be consistent
  * preferably "back compatible" in sense - minimal changes in existing code
* For old code my editors are incompatible  - **mostly when tabs are mixed with spaces**
  * Code is in some cases was practically unreadable

For your point c I had to reformat the code (reformatting is not done by default) as I could not read it (I can post screenshot). Once the code was reformatted I could reapply diff with -w option, (as yous suggest?), but I do not thing is is good idea. Code will be again not readable  and this is not what I wanted.
Next time - I will  explicitly mention formatting changes in commit.

In case we can not make a revolution (with old code), I prefer to apply default style for new code.
I need to know which style to apply:
*  K&R  ( as stated in https://en.wikipedia.org/wiki/Indent_style)
  *  most consistent with what we have currently in AliRoot and Root.
* how many spaces ?
  * in AliRoot we have mix of 2 resp. 4 space (sometimes tabs)

PR #301, opened 2017-07-01
Possibility to add offset in R,Z per particle

shahor02:
By calling AliGenBox::SetRandomOffset(rmin,rmax,zmin,zmax) each generated track will be forced to
have random offset within requested range (radial shift along phi of momentum). Needed for injection
of loopers in the TPC


shahor02:
+1

PR #302, opened 2017-07-02
STARLIGHT+AD update

hcab14:
- AD: initialization of TClonesArray TTree branches fixed: a similar problem has been found before but one instance of TClonesArray initialization was not corrected then.

- STARLIGHT: update to the latest svn revision from hepforge: this improves the printout for the cross section calculation and should not change the generation of particles


rbertens:
+1

mbroz84:
+1

PR #303, opened 2017-07-03
Conversion from cm/c to s for production time.

amorsch:
This should solve the problem with the production time in EPOS productions.
https://alice.its.cern.ch/jira/browse/PWGMM-21


rbertens:
+1

PR #304, opened 2017-07-03
curl instead of wget is now used for caching 

sebaleh:


miranov25:
+1

miranov25:
all tests were successful ....

PR #305, opened 2017-07-03
Conversion moved to NextEvent() since NextParticle() can be called several times

amorsch:
Conversion moved to NextEvent() since NextParticle() can be called several times

for the same particle, which leads to multiple application of the conversion factor.
Related to 
https://alice.its.cern.ch/jira/browse/PWGMM-21
Difference in tracking efficinecy in LHC17a2a (EPOSLHC) and LHC17a2b


qgp:
+1

PR #315, opened 2017-07-07
TClonesArray::Delete must be always called

hcab14:
AD/Calib/Saturation contains a `TTree` with `TClonesArray` branches where the `TClonesArray`s contain `TF1`s:

- When `TClonesArray::Delete() `is not called after `TTree::GetEntry()`, in the next call of `TTree::GetEntry()` `TF1`s with new names are added to the global list of functions **which always have the same address**
- If this happens more than once, the global list of functions contains several entries with the same address which leads to problems when the global list of functions is deleted.


ecalvovi:
+1

PR #318, opened 2017-07-10
Caching trees now done with curl

sebaleh:


miranov25:
+1

Modifications approved during hands-on session today

PR #320, opened 2017-07-10
Create on demand extra tree with vector of NTPC clusters per MC track

shahor02:
With AliTPCReconstructor::SetCountMCTrackClusters(kTRUE) static call
from rec.C a nclTPCperMCtrack.root file will be created with
srd::vector<short> containing number of TPC clusters per MC track. The full
content is:
nclPerMCtrack->Print()
...
*Br    0 :nMCTracks : nMCTracks/I                                            *
*Br    1 :nclTot    : nclTot/I                                               *
*Br    2 :nclOrphan : nclOrphan/I                                            *
*Br    3 :nclPerTrack : vector<short>                                        *
Since the same clusters may mention up to 3 MC tracks, the sum of Ncl per MC
track will normally exceed the total number of clusters.
nclOrphan gives number of clusters not originated by any track (noise)


shahor02:
+1

shahor02:
@davidrohr : here is the patch to extract NTPCcl per MC track to separate tree

PR #328, opened 2017-07-18
Updating C side compensator yoke definition to avoid overlaps

coppedis:


coppedis:
I looked at the log, but I don't see how the change of 3 cm in a magnet length described in the ZDC geometry can cause a crash in GRP object not found for the TPC. Can somebody help me in understanding, please?!?
Anyway I run runtest.sh locally and it finished without errors.

ch

ecalvovi:
It looks like the G4 sim automatic test went OK. Maybe it was a random crash not related to your changes. Can the test be redone?

PR #329, opened 2017-07-19
Introduce SetXYZcascade for testing purposes

ddobrigk:
This commit introduces a way to change the decay position estimate for AliESDcascades, as discussed with the class author, @iouribelikov . 

At the moment, this decay position is calculated in the standard AliESDcascade constructor via a simple 50/50 average of the points of closest approach of bachelor and V0. This is different wrt to what is done in V0s, in which the decay position is estimated using a weighted (with errors) average of the DCA points of negative and positive tracks, and we are currently exploring the possibility of using a similar approach for cascades. In offline checks, such an approach has been shown to result in an improvement in CosPA resolution of up to 15%, and the purpose of having a SetXYZcascade function is for us to be able to cross-check these offline checks in full-fledged grid analysis before moving to a more definite change of the AliESDcascade constructor.

### DIFF
--- a/STEER/ESD/AliESDcascade.cxx
+++ b/STEER/ESD/AliESDcascade.cxx
@@ -368,6 +368,15 @@ void AliESDcascade::GetXYZcascade(Double_t &x, Double_t &y, Double_t &z) const {
   z=fPosXi[2];
 }
 
+void AliESDcascade::SetXYZcascade(Double_t x, Double_t y, Double_t z) {
+    //--------------------------------------------------------------------
+    // This function sets cascade position (global)
+    //--------------------------------------------------------------------
+    fPosXi[0]=x;
+    fPosXi[1]=y;
+    fPosXi[2]=z;
+}
+
 Double_t AliESDcascade::GetDcascade(Double_t x0, Double_t y0, Double_t z0) const {
   //--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


dberzano:
+1 Thanks @iouribelikov for the thumbs up!

ddobrigk:
@dberzano , sorry for pinging so explicitly on this, but do we maybe have an ETA for this to appear in an AliRoot tag? I know that I unfortunately missed yesterday's tag, but I was really hoping to start using this simple SetXYZcascade function in our analyses soon.... Super thanks!

dberzano:
Hi @ddobrigk, as usual, we collect many requests before tagging, and there is a validation in between. We've just released yesterday. If we start a validation now, feedback will arrive during the weekend, therefore a new tag will not be done before Monday.

Open a JIRA ("Porting request" in AliRoot, use v5-09-12 as "Fix version(s)") and its urgency will be discussed next week at the CB.

ddobrigk:
Hi @dberzano , the jira ticket has [now been created](https://alice.its.cern.ch/jira/browse/ALIROOT-7397). I've noticed that there's been an AliRoot tag roughly every week or so recently, and while this is not extremely critical, I was really hoping to have this change in place for the next tag. Let's see. Thanks a lot!

PR #331, opened 2017-07-20
PWGPP-350 - ADD alihadd.sh and alihadd.test.sh

miranov25:
* function testFileList()
  * check root files
  * test read/write/delete
  * test only for selected key - array of regulr expression e.g
    * -checkKeys highPt - check highPt keys
    * -checkKeys . - check all keys
* function makeFilteredList()
  * to prepare lists of good and bad files
* function alihaddsh
  * to merge root files
    * using internaly alihadd
    * optional switch to checkKyes
```
alihaddsh -k -s 200000000   -checkKeys highPt -prefetch 1 -v 1 -timeOut 30 test.root @input.list
```
* test/alihadd.test.sh
  * test of alihadd.sh
  * some modification needed (input list) before integrating into ctest
    * most probably data from EOS/http to be used for test

### DIFF
--- a/STEER/Utilities/alihadd.sh
+++ b/STEER/Utilities/alihadd.sh
@@ -1,133 +1,123 @@
-#!/usr/bin/env bash
-# source $ALICE_ROOT/libexec/alihadd.sh
+#!/bin/bash
+
 # Example usage:
+#
 #   source $ALICE_ROOT/libexec/alihadd.sh
-#    * Filter input list before merging
-#       makeFilteredList input.list highPt filter.log
-#    *  Internal function
-#       testFileList input.list highPt | tee  filter.log
+#   # Filter input list before merging
+#   makeFilteredList input.list highPt filter.log
+#   # Internal function
+#   testFileList input.list highPt | tee  filter.log
 
-
-loadUtilities(){
-    #
-    #  load utilities if not yet done
-    if [ -z `type -t alilog_info` ]; then
-        source $ALICE_ROOT/libexec/alilog4bash.sh
-        alilog_info "Load ALICE_ROOT/libexec/alilog4bash.sh"
-    fi;
-    if [  -z `type -t guessPeriod` ]; then
-        source $ALICE_ROOT/libexec/utilities.sh
-        alilog_info "Load $ALICE_ROOT/libexec/utilities.sh"
-    fi;
+loadUtilities() {
+  #  load utilities if not yet done
+  if [[ -z $(type -t alilog_info) ]]; then
+    source $ALICE_ROOT/libexec/alilog4bash.sh
+    alilog_info "Load ALICE_ROOT/libexec/alilog4bash.sh"
+  fi
+  if [[ -z $(type -t guessPeriod) ]]; then
+    source $ALICE_ROOT/libexec/utilities.sh
+    alilog_info "Load $ALICE_ROOT/libexec/utilities.sh"
+  fi
 }
 
-testFileList(){
-    # Internal function: filter input list
-    # Provide good/bad file list out of inputList(param[$1])
-    # Tests: read/write/delete/eventually copytree for selected keys(param[$2])
-    inputList=$1
-    keyList=$2
-    timeOut=$3
-    alilog_info "inputList=$inputList"
-    alilog_info "keyList=$keyList"
-    for a in `cat $inputList`; do
+testFileList() {
+  # Internal function: filter input list
+  # Provide good/bad file list out of inputList(param[$1])
+  # Tests: read/write/delete/eventually copytree for selected keys(param[$2])
+  inputList=$1
+  keyList=$2
+  timeOut=$3
+  [[ $timeOut > 0 ]] && prefix="timeout $timeOut" || prefix=
+  alilog_info "inputList=$inputList"
+  alilog_info "keyList=$keyList"
+  for a in $(cat $inputList); do
     alilog_info "Filter.Start $a $keyList $timeOut"
-    prefix=""
-    if [ $timeOut > 0 ] ; then
-        prefix="timeout $timeOut"
-    fi;
-    export ROOT_HIST=0;
-    $prefix aliroot -n -b   -l   << EOF
+    ( export ROOT_HIST=0
+      $prefix aliroot -n -b -l <<EOF
 AliXRDPROOFtoolkit::TestFile("$a", "$keyList");
 EOF
+    )
     alilog_info "Filter.End $a  $?"
-    done
+  done
 }
 
-processFilterLog(){
-    # Internal function: parse log file produced by testFileList function
-    inputList=$1
-    logFile=$2
-     alilog_info "processFilterLog $logFile"
-     # get stat
-     ninput=`cat $logFile | grep -c "Filter.Start "`
-     nall=`cat $logFile | grep  -c "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_*"`
-     ngood=`cat $logFile | grep  -c -e "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_0"`;
-     alilog_info "processFilterLog $ninput:$nall:$ngood"
-    cat $logFile | grep "TestStatusAll_0" |gawk '{print $2}' >$inputList.Good
-    cat $logFile | grep "TestStatusAll_" | grep -v "TestStatusAll_0" |gawk '{print $2}' >$inputList.Bad
+processFilterLog() {
+  # Internal function: parse log file produced by testFileList function
+  inputList=$1
+  logFile=$2
+  alilog_info "processFilterLog $logFile"
+  # get stat
+  ninput=$(cat $logFile | grep -c "Filter.Start ")
+  nall=$(cat $logFile | grep  -c "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_*")
+  ngood=$(cat $logFile | grep  -c -e "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_0")
+  alilog_info "processFilterLog $ninput:$nall:$ngood"
+  cat $logFile | grep "TestStatusAll_0" | awk '{print $2}' > $inputList.Good
+  cat $logFile | grep "TestStatusAll_" | grep -v "TestStatusAll_0" | awk '{print $2}' > $inputList.Bad
 }
 
-makeFilteredList(){
-    if [ $# -lt 3 ]; then
-        printf "\n"
-        alilog_error "makeFilteredList.InvalidParameters $@"
-        printf "\n"
-        printf "Usage:  makeFilteredList inputList keyList logfile <timeOut>\n"
-        return
-    fi;
-    inputList=$1
-    keyList=$2
-    logFile=$3
-    timeOut=$4
-    testFileList $1 $2 $timeOut|tee $logFile
-    processFilterLog $1 $logFile
+makeFilteredList() {
+  if [ $# -lt 3 ]; then
+    printf "\n"
+    alilog_error "makeFilteredList.InvalidParameters $@"
+    printf "\n"
+    printf "Usage: makeFilteredList inputList keyList logfile <timeOut>\n"
+    return
+  fi
+  inputList=$1
+  keyList=$2
+  logFile=$3
+  timeOut=$4
+  testFileList $inputList $keyList $timeOut | tee $logFile
+  processFilterLog $inputList $logFile
 }
 
-alihaddsh(){
-    # the same as alihadd but additional flags
-    #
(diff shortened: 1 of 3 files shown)


miranov25:
Hallo @dberzano, @pzhristov
as we discussed before I included alihadd.sh and corresponding alihadd.test.sh into the AliRoot.
I prefer to review the code together  in person before actually getiing it into aliroot

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
```
dberzano: `#!/bin/bash`

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+#    * Filter input list before merging
+#       makeFilteredList input.list highPt filter.log
+#    *  Internal function
+#       testFileList input.list highPt | tee  filter.log
+
+
+loadUtilities(){
+    #
+    #  load utilities if not yet done
+    if [ -z `type -t alilog_info` ]; then
```
dberzano: Can you use:
* `[[` and `]]` (as opposed to `[` and `]`)
* `$(command)` (as opposed to backticks)

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+        source $ALICE_ROOT/libexec/utilities.sh
+        alilog_info "Load $ALICE_ROOT/libexec/utilities.sh"
+    fi;
+}
+
+testFileList(){
+    # Internal function: filter input list
+    # Provide good/bad file list out of inputList(param[$1])
+    # Tests: read/write/delete/eventually copytree for selected keys(param[$2])
+    inputList=$1
```
dberzano: Can you use uppercase names for shell variables (albeit local)?

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    alilog_info "inputList=$inputList"
+    alilog_info "keyList=$keyList"
+    for a in `cat $inputList`; do
+    alilog_info "Filter.Start $a $keyList $timeOut"
+    prefix=""
+    if [ $timeOut > 0 ] ; then
+        prefix="timeout $timeOut"
+    fi;
+    export ROOT_HIST=0;
+    $prefix aliroot -n -b   -l   << EOF
```
dberzano: I like the timeout thingy. Can you adjust the extra spaces?

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+AliXRDPROOFtoolkit::TestFile("$a", "$keyList");
+EOF
+    alilog_info "Filter.End $a  $?"
+    done
+}
+
+processFilterLog(){
+    # Internal function: parse log file produced by testFileList function
```
dberzano: Can you indent using two spaces?

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    inputList=$1
+    logFile=$2
+     alilog_info "processFilterLog $logFile"
+     # get stat
+     ninput=`cat $logFile | grep -c "Filter.Start "`
+     nall=`cat $logFile | grep  -c "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_*"`
+     ngood=`cat $logFile | grep  -c -e "I-AliXRDPROOFtoolkit::TestFile.TestAll:.*TestStatusAll_0"`;
+     alilog_info "processFilterLog $ninput:$nall:$ngood"
+    cat $logFile | grep "TestStatusAll_0" |gawk '{print $2}' >$inputList.Good
```
dberzano: Fix spaces. Use `awk` not `gawk`.

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    cat $logFile | grep "TestStatusAll_" | grep -v "TestStatusAll_0" |gawk '{print $2}' >$inputList.Bad
+}
+
+makeFilteredList(){
+    if [ $# -lt 3 ]; then
+        printf "\n"
+        alilog_error "makeFilteredList.InvalidParameters $@"
+        printf "\n"
+        printf "Usage:  makeFilteredList inputList keyList logfile <timeOut>\n"
```
dberzano: Spaces

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+        return
+    fi;
+    inputList=$1
+    keyList=$2
+    logFile=$3
+    timeOut=$4
+    testFileList $1 $2 $timeOut|tee $logFile
```
dberzano: Spaces before and after `|`

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    processFilterLog $1 $logFile
+}
+
+alihaddsh(){
```
dberzano: `alihaddsh()<space>{`

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    # the same as alihadd but additional flags
+    #
+    # Question  - hot to parse flags
+    # - checkKeys pattern
+    # - prefetch prefix
+    # source $ALICE_ROOT/libexec/alihadd.sh
+    # alihaddsh -k -s 1000000000   -checkKeys highPt -prefetch 1 -v 1 -timeOut 300 test.root @input.list
```
dberzano: Spaces

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    # alihaddsh -k -s 1000000000   -checkKeys highPt -prefetch 1 -v 1 -timeOut 300 test.root -inputList input.list
+    alilog_info "alihaddsh $*"
+    config="$@"
+    parseConfig $config
+    # check @inputList in parameter string if not specified as   -inputList
+    [ -z "$inputList" ] && inputList=`echo $config|grep " @"| sed -e 's/.*@\(.*\)/\1/'| gawk '{print $1}'`  #
```
dberzano: Dangling comment

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    if [ -z "$inputList" ]; then
+        alilog_error "Input list not specified"
+        exit 1;
```
dberzano: Useless `;`

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+    else
+        alilog_error "alihadd.sh $config"
+        return 1;
+    fi
+}
+
+main()
+{
+    loadUtilities;
+    eval "$@"
```
dberzano: I guess `$@` is sufficient and you might not need `eval`

STEER/Utilities/alihadd.sh
```diff
@@ -0,0 +1,133 @@
...
+}
+
+main $@
+
+#loadUtilities;
```
dberzano: Remove

STEER/Utilities/alihadd.sh
miranov25: is this generic
Clion suggest to use 
+#!/usr/bin/env bash

STEER/Utilities/alihadd.sh
miranov25: if [[ $(type -t alilog_info ]]
Is it problem of style or is usage of $() saver

I used  code above as it was one suggestion on stackoverflow

STEER/Utilities/alihadd.sh
miranov25: Until now I was using lower case variables.
In case you prefer some convention, I suggest to write list of suggestion 
per bash scripting and bash testing.
Short text document should be sufficient

Usually when I did not know how to implement something  - I consulted stack-overflow.
Of course people were using there different styles...

STEER/Utilities/alihadd.sh
miranov25: Not sure I understood your comment.
I had to use timeout because sometimes file tests ends were stuck.
I was debugging what happenned - some locks sets - within ALICE error handler
(I can find in JIRA ticket) in case of exceptions. 
timeout was the only option which I found

STEER/Utilities/alihadd.sh
miranov25: This should be in recommendation.
I can adjust CLion editor to predefined style. 
I will check consitency

STEER/Utilities/alihadd.sh
miranov25: OK

STEER/Utilities/alihadd.sh
dberzano: Use `#!/bin/bash` everywhere. We want it to be compatible with the system's one, not a generic bash.

STEER/Utilities/alihadd.sh
dberzano: Please comply with the bashism: double square braces and `$(...)`. The rest is for POSIX compliance but we do not care about that as we use bash, which is a better standard since 15 years.

STEER/Utilities/alihadd.sh
dberzano: Can you still use uppercase variables?

STEER/Utilities/alihadd.sh
dberzano: ` $prefix aliroot -n -b   -l   << EOF` should be `$prefix aliroot -n -b -l <<EOF`, I am referring to those extra in-between spaces between parameters.

STEER/Utilities/alihadd.sh
dberzano: I think two spaces are better, four are too many, thanks for complying!

dberzano:
Hi @miranov25, you can check my commit which has minor modifications. Given that this is `STEER` I am however wondering what to do with `alihadd.test.sh`. It contains batch code which seems to be quite specific to your use case and not portable: therefore it should just not be in AliRoot. _If_ you really need to have it committed, I suggest you place it somewhere else and outside `STEER`. What do you think?

miranov25:
Hallo @dberzano.

I checked your changes and they are following your and google bash coding convention.
I agree with this convention and I will try to use these convention in my further code.

Concerning alihadd.test.sh the test should be independent of the system.
The only specific part is the access to the reference data and setting of the some varaibles to submit jobs:
```
configMI(){
    source $ALICE_ROOT/libexec/alihadd.sh
    export prefixList=$NOTES/JIRA/PWGPP-340/testlists/
    batchCommand="/usr/bin/sbatch"
    batchFlags="--get-user-env --mem-per-cpu=6096 --time=8:00:00"
}
```
For sure I would like to run the test  as part of the integration test, specifying the input data - locating
in some predefined place. 
We should solve the problem of the input data location in some generic way - independent of the CTest.
In case you prefer outer directory, we can move the test into the 
$ALICE_ROOT/test/
But until  now I followed convention, test to be in sub-directory where the corresponding source code is.

Test code:
* local test          - should run everywhere
* batch test        - specific to nodes with batch system
* makeflow  test - should run everywhere once finished
 

Marian.

dberzano:
+test

Thanks for your explanation, it makes sense and I am going to approve it anyhow, as it is. It just does not make sense to have:

* local test
* batch test
* Makeflow test

when you can just have Makeflow test that covers all cases at once. We have learned that lesson when doing alisw/release-validation#28.

Well, you can as a matter of fact have a local test and a Makeflow one which are independent, to cover the case where Makeflow is not installed on the system. But seeing `sbatch` et al. upstream, when the use case is clearly the author's only, is IMHO wrong :-)

miranov25:
Hallo @dberzano 

Part of test will run within continuous test CTEST, some slower test within integration test.
I put the test code to the one shell script. 

Explanation of tests:
* local test - will be running within short CTEST - O(20 s)
  * basic test
* batch test - deprecated - will be removed once make-flow test will be running
  * I wanted to keep this test  until make-flow test will be working properly
* makeflow test  10xO(2 min)
  * should be run within integration test
  * more data/use cases to be tested
  * will replace obsolete batch-test

Anyway, we should solve the problem how to parameterize/configure input data location
and some specific setting.
E.g as we have in the  $ALICE_PHYSICS/PWGPP/benchmark/benchmark.config
```
#autoOCDB=0
defaultOCDB="local:///cvmfs/alice-ocdb.cern.ch/calibration/data/2010/OCDB/"
## makeflow options
#makeflowOptions="-T wq -N alice -C ali-copilot.cern.ch:9097"
makeflowOptions="-T condor"
```
Marian

dberzano:
+1
-- 
--
Dario Berzano
about.me/dario.berzano

PR #334, opened 2017-07-24
Fix OB Stave Length to avoid extrusions

mario6829:
The length of Outer Barrel Staves was wrongly computed in CreateITSUv2 macro. Fixed to match the actual Module length and so avoid volume extrusions in geometry.


pzhristov:
+test

pzhristov:
+1

PR #335, opened 2017-07-24
fix throw usage in gcc7+root6 case for STEER/STEER/AliAlignObj*

adriansev:


aaniin:
Hello @dberzano @ktf @shahor02,

Can you give this PR a +1 please? In c++11 the exception specifier was deprecated which in combination with -Werror leads to compilation errors. This PR replaces `throw "foo"` with `AliFatal("foo")`. 

Cheers,
Hans

shahor02:
+1

PR #339, opened 2017-07-28
Avoid overlap with ZEM in corrected position.

amorsch:
AliZDCv4 will be updated with corrected ZEM position.
To avoid overlaps the dimensions and position of voRB24CuTubeA have been changed.


dberzano:
+1

PR #341, opened 2017-07-31
Basic documentation of the interface of the calo trigger object

mfasDa:
Where available information of the scope (per patch, per fastor)
added, plus some basic information about the return types. Not
yet complete, more information to be added.


shahor02:
+1

PR #343, opened 2017-08-04
AliMagFast: Extend solenoid Z region to -550:550cm, and implemented 22…

syamasak:
… uniform Z bins.

Note that the max deviations of the TOF Bz out of z6:16z are >100%, especially z0,1,19,20,21 regions have max >1k% deviation.


shahor02:
Thanks @syamasak , I'll check it 1st locally and if ok will merge the PR

shahor02:
+1

PR #344, opened 2017-08-07
Changes to do with primary particle definition as implemented in AliStack::IsPhysicalPrimary

cholmcc:
- The code of `AliStack::IsPhysicalPrimary` was changed to conform with
  the definition adopted by ALICE in a recent public note:

    A primary particle is a particle with a mean proper lifetime
    tau larger than 1cm/c, which is either a) produced directly in the
    interaction, or b) from decays of particles with tau smaller
    than 1cm/c, restricted to decay chains leading to the interaction.

  See also ALICE-PUBLIC-2017-005 at

     https://cds.cern.ch/record/2270008

  Please note, that the definition is approved by the Physics Board

  - This is the modified implementation which stops the search for long-lived mothers 
    when encountering beam particles (status code 3 or 4).  It also ignores particles flagged
    by the EG as intermittent. 

- The old implementation of `AliStack::IsPhysicalPrimary` is still
  available as `AliStack::IsPhysicalPrimaryOld` to allow testing etc.

- Various bugs in the old implementation (now `AliStack::IsPhysicalPrimaryOld`)
  and in `AliStack::IsSecondaryFromWeakDecay` were corrected.  Specifically
  these bugs had to do with extracting the heaviest flavour or a
  particle (to see if a particle was heavy flavour or strange)

  To get the heaviest flavour of a particle with PID pdg the old code
  tried to extract the heaviest flavour quark by doing

    `f = floor(|pdg| / 10^{floor(log10(|pdg|)})`

  Value of f corresponds to the heaviest quark in the particle
  (1: d, 2: u, 3: s, 4: c, 5: b, 6; t) For most particles
  this is fine, e.g.,

  - pi+:   pdg=211  -> f=2
  - K+:    pdg=311  -> f=3
  - K0(S): pdg=310  -> f=3
  - K0(L): pdg=130  -> f=1(!)
  - p      pdg=2212 -> f=2
  - B+:    pdg=521  -> f=5
  - D+:    pdg=411  -> f=4

  while for others it will give the wrong result

  - D(0)*      pdg=10421   -> f=1 (should be 4)
  - K(1)+      pdg=10323   -> f=1 (should be 3)
  - f(0)(500)  pdg=9000221 -> f=9(!) (should be 2)

  and the leptons are really messed up

  - e-   pdg=11 -> f=1
  - mu-  pdg=13 -> f=1
  - tau- pdg=15 -> f=1

  The definition of the PDG code conventions state that each
  PDG code is formatted as

    `n nr nL nq1 nq2 nq3 nJ`

  where `nq1`, `nq2`, and `nq3` encode the quark content.
  For mesons `nq1=0`.  So a better way to extract the heaviest
  flavour is to take the absolute value of the PDG code, modulo
  10,000 and then, depending on whether the value is larger than
  999 (baryons) extract the 4th or 3rd digit

```
    if (|pdg|<100 && |pdg|>6) return 0; // No quark content
    m = |pdg| % 10000; // get nq1 nq2 nq3 nJ
    if (m == K(0)L) return 3; // special for K0L
    return (m > 999) ? m / 1000 : m/100;
```

  Note, for nuclei the convention is to encode as

    `10LZZZAAAI`

  where `L` is the number of Lambda's in the nucleus.  So we
  should check if this number is non-zero and in that case
  get 3 (strange), otherwise 2 (up).  I.e., the full
  pseudo-code is

    ```
    if (|pdg|>1000000000)
      f = (|pdg|/10000000) % 10) != 0 ? 3 : 2;
    else if (|pdg|<100 && |pdg|>6)
      f = 0; // No quark content
    else
      m = |pdg| % 10000; // get nq1 nq2 nq3 nJ
      if (m == K(0)L) return 3; // special for K0L
      f = (m > 999) ? m / 1000 : m/100;
```

  This algorithm is implemented in the service function
  `HeaviestFlavour` in the scope of **AliStack.cxx**

- Other small efficiency fixes.

### DIFF
--- a/STEER/STEERBase/AliStack.cxx
+++ b/STEER/STEERBase/AliStack.cxx
@@ -983,6 +983,80 @@ Bool_t AliStack::GetEvent()
     ResetArrays(size);
     return kTRUE;
 }
+
+// Some utilities used 
+namespace {
+    /** 
+   * Check the heaviest flavour of a particle.  Note, the original
+   * implementation is
+   *
+   *@f$f =  \left\lfloor\frac{c}{10^{\lfloor\log_{10}c\rfloor}}\right\rfloor@f$
+   * 
+   * since it will return 1 for @f$\Upsilon(2\mathrm{S})@f$ (@f$ c=100553@f$). 
+   * 
+   * The coding convention for PDG numbers is 
+   *
+   * @f$ n n_r n_L n_{q1} n_{q2} n_{q3} n_J@f$ 
+   *
+   * which means to guage the quark content we need only the 4
+   * right-most digits.  The most optimal way to extract these is to
+   * format the modulo of the passed number to @f$10\,000@f$, format
+   * that remainder as a string and then take the first character and
+   * convert that to a number by subtracting off the character @c '0'.
+   * 
+   * @param pdg 
+   * 
+   * @return 
+   */
+  Int_t HeaviestFlavour(Int_t pdg)
+  {
+    Int_t apdg = TMath::Abs(pdg);
+    // Special codes for bosons and leptons 
+    if (apdg < 100 && apdg > 6) return 0;
+    // Special codes for nuclei
+    // Ion/nucleus with format 10LZZZAAAI - take the 10L part modulo
+    // 10 to get the L part.  Hyper nucleons with np protons, nn
+    // neutrons, and nl Lambdas have
+    //
+    //    AAA = np+nn
+    //    ZZZ = np
+    //    L   = nl
+    //    I   = isomer level
+    //
+    // Deutron: 1000010020
+    // U(235):  1000922350
+    // 
+    if (apdg > 1000000000) return ((apdg/10000000) % 10) != 0 ? kStrange : kUp;
+    // Take lower 4 digits 
+    Int_t mpdg = apdg % 10000;
+    // Special case for K0L
+    if (mpdg == kK0Long) return kStrange;
+    // If a baryon (>999) then the 4th digit, other wise the 3rd 
+    return mpdg > 999 ? mpdg/1000 : mpdg / 100;
+  }
+  /** 
+   * Check if a particle is heavy flavoured 
+   * 
+   * @param pdg Particle identification code 
+   * 
+   * @return True if the heaviest quark is a charm, bottom, or top
+   */
+  Bool_t IsHeavyFlavour(Int_t pdg)
+  {
+    return HeaviestFlavour(pdg) >= kCharm;
+  }
+  /** 
+   * Check if particle is strange 
+   * 
+   * @param pdg Particle identification code 
+   * 
+   * @return true if heaviest quark is a strange 
+   */
+  Bool_t IsStrange(Int_t pdg)
+  {
+    return HeaviestFlavour(pdg) == kStrange;
+  }
+}
 //_____________________________________________________________________________
 
 Bool_t AliStack::IsStable(Int_t pdg) const
@@ -990,49 +1064,39 @@ Bool_t AliStack::IsStable(Int_t pdg) const
   //
   // Decide whether particle (pdg) is stable
   //
-  
-  
+  Int_t apdg = TMath::Abs(pdg);
+    		    
   // All ions/nucleons are considered as stable
   // Nuclear code is 10LZZZAAAI
-  if(pdg>1000000000)return kTRUE;
+  if(apdg>1000000000)return kTRUE;
 
-  const Int_t kNstable = 18;
-  Int_t i;
-  
-  Int_t pdgStable[kNstable] = {
-    kGamma,             // Photon
-    kElectron,          // Electron
-    kMuonPlus,          // Muon 
-    kPiPlus,            // Pion
-    kKPlus,             // Kaon
-    kK0Short,           // K0s
-    kK0Long,            // K0l
-    kProton,            // Proton 
-    kNeutron,           // Neutron
-    kLambda0,           // Lambda_0
-    kSigmaMinus,        // Sigma Minus
-    kSigmaPlus,         // Sigma Plus
-    3312,               // Xsi Minus 
-    3322,               // Xsi 
-    3334,               // Omega
-    kNuE,               // Electron Neutrino 
-    kNuMu,              // Muon Neutrino
-    kNuTau              // Tau Neutrino
-  };
-    
-  Bool_t isStable = kFALSE;
-  for (i = 0; i < kNstable; i++) {
-    if (pdg == TMath::Abs(pdgStable[i])) {
-      isStable = kTRUE;
-      break;
-    }
+  switch (apdg) {
+  case kGamma:             // 22   Photon
+  case kElectron:          // 11   Electron
+  case kMuonMinus:         // 13   Muon 
+  case kPiPlus:            // 211  Pion
+  case kKPlus:             // 321  Kaon
+  case kK0Short:           // 310  K0s
+  case kK0Long:            // 130  K0l
+  case kProton:            // 2212 Proton 
+  case kNeutron:           // 2112 Neutron
+  case kLambda0:           // 3122 Lambda_0
+  case kSigmaMinus:        // 3112 Sigma Minus
+  case kSigmaPlus:         // 3222 Sigma Plus
+  case kXiMinus:           // 3312 Xi Minus 
+  case 3322:               //      Xi 0 
+  case kOmegaMinus:        // 3334 Omega
+  case kNuE:               // 12   Electron Neutrino 
+  case kNuMu:              // 14   Muon Neutrino
+  case kNuTau:             // 16   Tau Neutrino
+    return true;
   }
-  
-  return isStable;
+  return false;
 }
 
+
 //_____________________________________________________________________________
-Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
+Bool_t AliStack::IsPhysicalPrimaryOld(Int_t index, Bool_t useInEmbedding)
 {
     //
     // Test if a particle is a physical primary according to the following definition:
@@ -1044,7 +1108,10 @@ Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
     Int_t ist = p->GetStatusCode();
     
     //
-    // Initial state particle
+    // Initial state particle - note, this is tricky - it depends on
+    // what the EG defines as final state (status=1) particles.  Could
+    // be very late in the game if particles are allowed to decay late
+    // in the EG.
     if (ist > 1) return kFALSE;
     
     Int_t pdg = TMath::Abs(p->GetPdgCode());
@@ -1071,10 +1138,10 @@ Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
 	if ((mpdg == kPi0) && (imo < GetNprimary()))   return kTRUE; 
 
 // Check if this is a heavy flavor decay product
-	Int_t mfl  = Int_t (mpdg / TMath::Power(10, Int_t(TMath::Log10(mpdg))));
+	Int_t mfl  =  HeaviestFlavour(mpdg);
 	//
 	// Light hadron
-	if (mfl < 4) return kFALSE;
+	if (mfl < kCharm) return kFALSE;
 	
 	//
 	// Heavy flavor hadron produced by generator
@@ -1089,9 +1156,9 @@ Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
 	    pm  =  Particle(imo,useInEmbedding);
 	}
 	mpdg = TMath::Abs(pm->GetPdgCode());
-	mfl  = Int_t (mpdg / TMath::Power(10, Int_t(TMath::Log10(mpdg))));
+	mfl  = HeaviestFlavour(mpdg);
 
-	if (mfl < 4) {
+	if (mfl < kCharm) {
 	    return kFALSE;
 	} else {
 	    return kTRUE;
@@ -1099,6 +1166,82 @@ Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
     } // produced by generator ?
 } 
 
+//_____________________________________________________________________________
+Bool_t AliStack::IsPhysicalPrimaryNew(Int_t index, Bool_t useInEmbedding)
+{
+  // Check wether a given particle is a primary according to the
+  // definition adopted by ALICE:
+  //
+  //   A primary particle is a particle with a mean proper lifetime
+  //   tau than 1cm/c, which is either a) produced directly in the
+  //   interaction, or b) from decays of particles with tau smaller
+  //   than 1cm/c, restricted to decay chains leading to the interaction.
+  //
+  // See also ALICE-PUBLIC-2017-005 at
+  //
+  //   https://cds.cern.ch/record/2270008
+  //
+  // for more on this
+  TParticle* p = Particle(index,useInEmbedding);
+
+  // Check if we should ignore this particle - i.e., all particles
+  // that the EG has flagged as intermiddent particles.
+  Int_t ist = p->GetStatusCode();
+  if (ist == 0 || (ist >= 11 && ist <= 200)) return false;
+  
+  // Check if this particle has tau >= 1cm/c 
+  if (!IsStable(p->GetPdgCode())) return false;
+
+  // Check if this comes from a primary process (i.e., decay or the
+  // generator)
+  switch (p->GetUniqueID()) {
+  case kPDecay:
+  case kPNoProcess:
+  case kPNull:
+  case kPPrimary:
+    break;
+  default:
+    return false;
+  }
+
+  // Loop back over all mothers
+  TParticle* m  = p;
+  Int_t      mi = 0;
+  while ((mi = m->GetFirstMother()) >= 0) {
+    // Get (grand)mother 
+    m = Particle(mi,useInEmbedding);
+    // Check if (grand)mother is beam particle.  Note, Pythia6 does
+    // not follow HepMC standard and uses status code 3 for initial
+    // state - sigh!
+    Int_t mist = m->GetStatusCode();
+    if (mist == 3 || mist == 4) return true;
+    
+    
+    // If (grand)mother has tau >= 1cm/c, then this (p) isn't a
+    // primary
+    if (IsStable(m->GetPdgCode())) return false;
+    
+    // Check that the (grand)mother comes from a primary process e.g.,
+    // decay or from the generator
+    switch (m->GetUniqueID()) {
+    case kPDecay:
+    case kPNoProcess:
+    case kPNull:
+    case kPPrimary:
+      break;
+    default:
+      return false;
+    }
+  }
+  return true;
+}
+//_____________________________________________________________________________
+Bool_t AliStack::IsPhysicalPrimary(Int_t index, Bool_t useInEmbedding)
+{
+  return IsPhysicalPrimaryNew(index, useInEmbedding);
+}
+
+//_____________________________________________________________________________
 Bool_t AliStack::IsSecondaryFromWeakDecay(Int_t index, Bool_t useInEmbedding) {
 
   // If a particle is not a physical primary, check if it comes from weak decay
@@ -1115,16 +1258,11 @@ Bool_t AliStack::IsSecondaryFromWeakDecay(Int_t index, Bool_t useInEmbedding) {
   // mass of the flavour
   Int_t mfl = 0;
   // Protect the "rootino" case when codemoth is 0
-  if (TMath::Abs(codemoth)>0) mfl = Int_t (codemoth / TMath::Power(10, Int_t(TMath::Log10(codemoth))));
+  if (TMath::Abs(codemoth)>0) mfl = HeaviestFlavour(codemoth);
   
-  if(mfl == 3 && uniqueID == kPDecay) return kTRUE;// The first mother is strange and it's a decay
-  if(codemoth == 211 && uniqueID == kPDecay) return kTRUE;// pion+- decay products
-  if(codemoth == 13 && uniqueID == kPDecay) return kTRUE;// muon decay products
-
-  /// Hypernuclei case
-  if (TMath::Abs(moth->GetPdgCode()) > 1000000000 && uniqueID == kPDecay) {
-    if ((moth->GetPdgCode() / 10000000) % 10 != 0) return kTRUE; /// Number of lambdas in the hypernucleus != 0
-  }
+  if(mfl == kStrange && uniqueID == kPDecay) return kTRUE;// The first mother is strange and it's a decay
+  if(codemoth == kPiPlus && uniqueID == kPDecay) return kTRUE;// pion+- decay products
+  if(codemoth == kMuonMinus && uniqueID == kPDecay) return kTRUE;// muon decay products
 
   return kFALSE;
   
--- a/STEER/STEERBase/AliStack.h
+++ b/STEER/STEERBase/AliStack.h
@@ -82,6 +82,8 @@ class AliStack : public TVirtualMCStack
     TParticle*  ParticleFromTreeK(Int_t id, Bool_t useInEmbedding=kFALSE) const;
     Int_t       TreeKEntry(Int_t id, Bool_t useInEmbedding=kFALSE) const;
     Bool_t      IsPhysicalPrimary(Int_t i, Bool_t useInEmbedding=kFALSE);
+    Bool_t      IsPhysicalPrimaryNew(Int_t i, Bool_t useInEmbedding=kFALSE);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by chiarazampolli** [comment]
Dear Christian,

As just discussed in the Computing Board, could you attach here a plot/printout that shows that the issue in the previous commit is fixed? In the other pull request https://github.com/alisw/AliRoot/pull/271#issuecomment-317037162 Andreas commented:

"The while-loop in AliStack::IsPhysicalPrimary can go back to the colliding particles which are stable.
In this case the particle is falsely flagged as a secondary. This happens quite frequently. We have to switch back to the old version of the code until this is fixed."

Maybe someone knows if there are more specific plots that could help with this. One plot that was sent around and showed that there were problems was the fraction of reconstructed primary pions, kaons, protons obtained with the old code and the new one. 

Thanks, and regards,

Chiara

**Comment by cholmcc** [comment]
Hi Chiara,

I asked Andreas - in a separate thread - if he could run his checks again
with the new fix. Let's see if he comes up with something - I guess he's
got it set up already.

Yours,

Christian Holm Christensen --------------------
Sankt Hans Gade 23 4, DK-2200 Copenhagen
+4524618591, cern.ch/cholm

Den 7. aug. 2017 16.36 skrev "chiarazampolli" <notifications@github.com>:

**Comment by chiarazampolli** [comment]
Thanks Christian! I know that Andreas is on holiday till 21 August, so he might respond with some delay. Maybe we could ask Nicolo' Jacazio to run his checks with this patch too? @dberzano , how do I find Nicolo's username? I thought it was njacazio, but the completion does not work...

Cheers,

Chiara

**Comment by cholmcc** [comment]
Hi Chiara, 

Nicolo's email seem to be 

`nicolo.jacazio@cern.ch
`

See https://greybook.cern.ch/greybook/participant/detail?id=762230

His GitHub handle seems to be 

`@njacazio 
`

See https://github.com/njacazio

Hope that helps. 

Yours,

Christian

**Comment by chiarazampolli** [comment]
Ciao Christian, 

Thanks! then I thought his account was correct, but github does not seem to complete. Anyway I will try now.

@njacazio, can you run your code that spotted the issue with the IsPhysicalPrimary with the changes that Christian put here?

Cheers,

Chiara

**Comment by njacazio** [comment]
Dear All,
I performed the requested checks. 
Essentially I applied last commit of Christian's repository to my local version of AliRoot and rerun the analysis.
You can find the root files with the comparison at the link https://cernbox.cern.ch/index.php/s/33erEy2N56YK5h9
Please note that the statistics for now is low.
Let me know if you have any suggestion!
Thanks,
Cheers,
Nicolo'

**Comment by cholmcc** [comment]
Hi Nicolò,




Den 8. aug. 2017 18.21 skrev "Nicolò Jacazio" <notifications@github.com>:

Dear All,
I performed the requested checks.
Essentially I applied last commit of Christian's repository to my local
version of AliRoot and rerun the analysis.
You can find the root files with the comparison at the link
https://cernbox.cern.ch/index.php/s/33erEy2N56YK5h9


I cannot access the share - says it's password protected, and I do not know
the password.


Please note that the statistics for now is low.
Let me know if you have any suggestion!


Could you summarize your findings in a few words?

Thanks.

Yours,

Christian Holm Christensen --------------------
Sankt Hans Gade 23 4, DK-2200 Copenhagen
+4524618591, cern.ch/cholm

**Comment by cholmcc** [comment]
Hi Nicolò et al, 

Thanks for the access. I got the files. I attach the three plots here for everybody's convenience 
k
p
pi

There are 4 sets of points in each plot 

- **OldVan** (blue) - which I assume is _before_ the new implementation
- **NewVan** (red) - which I assume is _after_ the new implementation, but _before_ the fix from yesterday
- **NewVan_OldMethod** (yellow) - which which I assume is _after_ bug-fixing the old implementation
- **Patch** (green) - which I assume is _after_ the new implementation _and_ the fix from yesterday

Some questions 

- The ordinate is clearly the transverse momentum, but what exactly is the abscissa?
- Does  a value of one for a given transverse momentum indicate no data for that data set and particle specie? 
- Which EG did you run?

Some observations

- It seems that the green points (**Patch**) come from a small number of events, and the statistical uncertainties are significant. 
- For protons, the fix (**Patch**) follows the old implementation (**OldVan**), and the bug-fixed old implementation agrees as well 
- For Kaons there's not enough data for to judge how the fix (**Patch**) fares, but the bug-fixed version of the old implementation agrees. 
- For pions, the fix (**Patch**) and bug-fixed old implementation agrees with the old implementation _within_ the statistical uncertainties. 

Would you agree with these observations? 

Yours, 

Christian

**Comment by njacazio** [comment]
Hello Christian,
sorry for not describing the plots and not attaching them directly to the thread.

- All plots indicate the **primary fraction of reconstructed tracks**, i.e. :
The distribution of  _reconstructed_ **primary** tracks (with MC truth on the PID) divided by the distribution of  _reconstructed_  tracks (with MC truth on the PID)
  - A value 1 (of the primary fraction) at a given pT indicates that the number of reconstructed tracks is equal to the number of reconstructed tracks labeled as `IsPhysicalPrimary` at that pT
- Track selection is the same as the one defined in `AliESDtrackCuts::GetStandardITSTPCTrackCuts2011 `

_________________________________________________________

* _OldVan_ : 
  - obtained with an AliPhysics version before the change in the `IsPhysicalPrimary` method
* _NewVan_ : 
  - obtained with an AliPhysics version after the change in the `IsPhysicalPrimary` method (i.e. the removed tags)
* _NewVan_OldMethod_ :
  - obtained with the same AliPhysics version of  NewVan but using the `IsPhysicalPrimaryOld` method
* _Patch_ : 
  -  the latest test, obtained by using the patch from your repository

_________________________________________________________

* My apologies, but what do you mean by EG?

* I agree with your observations:
  - Indeed the fixed version (_Patch_) seems to show agreement with the old (_OldVan_) `IsPhysicalPrimary` and (_NewVan_OldMethod_) `IsPhysicalPrimaryOld` methods
 
* As you said statistical fluctuations are quite large so I would like to know if there is the possibility to run a more conclusive test by performing the same check on grid and also with the (current) `IsPhysicalPrimary` method. I think that this would require a test version of AliRoot with the fix. If you agree I could also add a new curve with the current `IsPhysicalPrimaryOld`.

  -  In case you have other suggestions, please don't hesitate to ask!

Thanks,
Best regards,
Nicolo'

**Comment by cholmcc** [comment]
Hi Nicolò et al,

> All plots indicate the primary fraction of reconstructed tracks, i.e. :
> The distribution of reconstructed primary tracks (with MC truth on the PID) divided by the distribution of 
> reconstructed tracks (with MC truth on the PID)

Ah, so 

  ` N_{x,primary & reconstructed}(pt) / N_{x, reconstructed}(pt)   for x=pi,K,p
`
> A value 1 (of the primary fraction) at a given pT indicates that the number of reconstructed tracks 
> is equal to the number of reconstructed tracks labeled as IsPhysicalPrimary at that pT

So when, in the p+p&#772; case, where **Patch** is mostly unity but the others are not, I guess that's because of the limited sample size used, meaning few observations in those bins.

> My apologies, but what do you mean by EG?

I mean _Event_ _Generator_ - e.g., Pythia, Hijing, EPOS-LHC, DPMJet, ...

> As you said statistical fluctuations are quite large so I would like to know if there is the possibility to 
> run a more conclusive test by performing the same check on grid and also with the (current) 
> IsPhysicalPrimary method. I think that this would require a test version of AliRoot with the fix. 
> If you agree I could also add a new curve with the current IsPhysicalPrimaryOld.

In my patch I define 

- `IsPhysicalPrimaryNew` - which corresponds to the new definition 
- `IsPhysicalPrimaryOld` - which corresponds to the (bug-fixed) old implementation 

and 

- `IsPhysicalPrimary` calls `IsPhysicalPrimaryNew`.

If this latter function is redefined to call `IsPhysicalPrimaryOld`, then one could propagate the change to upstream AliROOT without breaking anything, but allow you to define a task that tests the two implementations against each other.  Then, once we're happy, we can change the implementation of `IsPhysicalPrimary` to call `IsPhysicalPrimaryNew` (or replace it entirely). 

Another option I've used, is to reimplement the methods in a task - e.g., 

```C++
struct TestPrimary : public AnalysisTaskSE
{
  Int_t HeaviestFlavour(Int_t pdg)
  {
    Int_t apdg = TMath::Abs(pdg);
    if (apdg > 1000000000) return ((apdg/10000000) % 10) != 0 ? kStrange : kUp;
    Int_t mpdg = apdg % 10000;
    if (mpdg == kK0Long) return kStrange;
    return mpdg > 999 ? mpdg/1000 : mpdg / 100;
  }
  Bool_t IsStable(Int_t pdg) const
  {
    Int_t apdg = TMath::Abs(pdg);
    if(apdg>1000000000)return kTRUE;

    switch (apdg) {
    case kGamma:             // 22   Photon
    case kElectron:          // 11   Electron
    case kMuonMinus:         // 13   Muon 
    case kPiPlus:            // 211  Pion
    case kKPlus:             // 321  Kaon
    case kK0Short:           // 310  K0s
    case kK0Long:            // 130  K0l
    case kProton:            // 2212 Proton 
    case kNeutron:           // 2112 Neutron
    case kLambda0:           // 3122 Lambda_0
    case kSigmaMinus:        // 3112 Sigma Minus
    case kSigmaPlus:         // 3222 Sigma Plus
    case kXiMinus:           // 3312 Xi Minus 
    case 3322:               //      Xi 0 
    case kOmegaMinus:        // 3334 Omega
    case kNuE:               // 12   Electron Neutrino 
    case kNuMu:              // 14   Muon Neutrino
    case kNuTau:             // 16   Tau Neutrino
      return true;
    }
    return false;
  }
  Bool_t IsPhysicalPrimaryOld(AliStack* stack, Int_t index, Bool_t useInEmbedding=false)
  {
    TParticle* p = stack->Particle(index,useInEmbedding);
     Int_t ist = p->GetStatusCode();
     if (ist > 1) return kFALSE;
     Int_t pdg = TMath::Abs(p->GetPdgCode());
     if (!IsStable(pdg)) return kFALSE;
     if (index < stack->GetNprimary()) return true;
     Int_t imo =  p->GetFirstMother();
     TParticle* pm  = stack->Particle(imo,useInEmbedding);
     Int_t mpdg = TMath::Abs(pm->GetPdgCode());
     if ((mpdg == 3212) &&  (imo <  stack->GetNprimary())) return kTRUE;
     if ((mpdg == kPi0) && (imo < stack->GetNprimary()))   return kTRUE; 
     Int_t mfl  =  HeaviestFlavour(mpdg);
     if (mfl < kCharm) return kFALSE;
     if (imo <  GetNprimary())  return kTRUE;
     while (imo >=  GetNprimary()) {
	imo = pm->GetFirstMother();
	pm  =  stack->Particle(imo,useInEmbedding);
    }
    mpdg = TMath::Abs(pm->GetPdgCode());
    mfl  = HeaviestFlavour(mpdg);
    if (mfl < kCharm)   return kFALSE;
    return kTRUE; 
  }
  Bool_t IsPhysicalPrimaryNew(AliStack* stack, Int_t index, Bool_t useInEmbedding)
  {
    TParticle* p = stack->Particle(index,useInEmbedding);
    Int_t ist = p->GetStatusCode();
    if (ist == 0 || (ist >= 11 && ist <= 200)) return false;
    if (!IsStable(p->GetPdgCode())) return false;
    switch (p->GetUniqueID()) {
    case kPDecay: case kPNoProcess: case kPNull: case kPPrimary: break;
    default: return false;
    }
    TParticle* m  = p;
    Int_t      mi = 0;
    while ((mi = m->GetFirstMother()) >= 0) {
      m = stack->Particle(mi,useInEmbedding);
      Int_t mist = m->GetStatusCode();
      if (mist == 3 || mist == 4) return true;
      if (IsStable(m->GetPdgCode())) return false;
      switch (m->GetUniqueID()) {
      case kPDecay: case kPNoProcess: case kPNull: case kPPrimary: break;
      default: return false;
      }
    }
    return true;
  }
  void UserExec(Option_t* option="")
  {
    AliMCEvent*  mcEvent = MCEvent();
    if (!mcEvent || !mcEvent->Stack()) return;
    AliStack* stack = mcEvent->Stack();
    Int_t nTracks   = stack->GetNtrack();
    Int_t nPrim     = stack->GetNprimary();
    for (Int_t i = 0; i < nTracks; i++) { 
      AliMCParticle* particle =
	static_cast<AliMCParticle*>(mcEvent->GetTrack(i));
      bool              isOldBug = stack->IsPhysicalPrimary(i);
      bool              isOldFix  = IsPhysicalPrimaryOld(stack, i);
      bool              isNew     = IsPhysicalPrimaryNew(stack, i);
       // Fill histograms here
    }
  }
};
```
With something like this, you do not need to have the patch applied to AliROOT.

I did something similar in my tests for Rivet (except I wrote out a simple tree that I converted to HepMC records) - see 

-  https://gitlab.cern.ch/cholm/alice-rivet/blob/master/tohepmc/README.md
-  http://alice-rivet.web.cern.ch/alice-rivet/class_digest_kinematics.html
-  http://alice-rivet.web.cern.ch/alice-rivet/class_digest_train.html

amorsch:
Hello Christian, all,
in the code I see the following comments 

// Check if (grand)mother is beam particle.  Note, Pythia6 does
// not follow HepMC standard and uses status code 3 for initial
// state - sigh!
//  Int_t mist = m->GetStatusCode();
//  if (mist == 3 || mist == 4) return true;

So if I understand correctly the code works only for generators that respect the HepMC convention ?
I tried with Pythia6 through the TPythia interface and it does not work, because the initial protons do not have status codes 3 or 4. So why should we have this code in production ?
Andreas

cholmcc:
Hi Andreas, 

The code should work for all EG's.  The check for the status code - whichs isn't commented out - is for the EG's (or TGenerator based interfaces) that export the full event history - such as Pythia6 and Pythia8. 

In my tests on standalone Pythia6 I find that the initial protons _always_ have status code 3.  This is easy to see by printing the event record. Now, of course, that event record comes about from 

```Fortran
      CALL PYEVNT
      CALL PYHEPC
C Export HEPEVT to HepMC
```
where `PYHEPC` does 

```Fortran
          ISTHEP(I)=0
          IF(K(I,1).GE.1.AND.K(I,1).LE.10) ISTHEP(I)=1
          IF(K(I,1).GE.11.AND.K(I,1).LE.20) ISTHEP(I)=2
          IF(K(I,1).GE.21.AND.K(I,1).LE.30) ISTHEP(I)=3
          IF(K(I,1).GE.31.AND.K(I,1).LE.100) ISTHEP(I)=K(I,1)
```
which indicates that the `PYJETS/K(*,1)` is 21 for initial state particles since that's te he only code used in the range from 21 to 30 (both inclusive - see 
page 60 of http://home.thep.lu.se/~torbjorn/pythia6/lutp0613man2.pdf).  So perhaps, the check should be for 21 (Pythia6) and 4 (HepMC standard).  

Other generators I've seen do not export the initial particles or give these codes not considered.

But clearly, we should check.  I do not know which EG Nicolò ran - but for that EG it seems to work.

Ideally, _either_ the EGs would give sensible status codes (i.e., according to e.g., the HepMC standard), _or_ the `TGenerator` derived classes would do the conversion.  The HepMC standard is pretty clear on the issue:

- 0 is ignored
- 1 means "final state" - i.e., _not_ decayed
- 2 means a particle that has decayed 
- 4 means initial (or beam) particle. 
- 11-100 are for MC specific codes 
- 200- are for auxiliary codes - e.g., transport by detector simulation

The status for AliROOT provided `TGenerator` interfaces are 

- `TAmpt` assumes  particles from the model are final state and has status=1 (and no history)
- `THerwig` uses the Pythia6 HEPEVT common block, and as such flags initial particles with status=3
- `TStarlight` when run with Pythia6 uses the same codes as TPythia6 (21 for initial state), and status code 0 for all particles in DPMJet mode (and no history)at
- `TDPMJet` uses status code 1 for final state, -1 for fragments of initial state, and 1001 for residual nucleus, but DPMJet does not include full history
- `TEpEmGen` always gives status code -1
- `AliGenEvtGen` changes the decayed particles status to 11 (sigh) and assigns status=1 to other particles 
- `THijing` gives status 1 for "final state", and 11 for decayed particles, and no history
- `THydJet` is like `TPythia6` - 1=final state, 11=decay, 21=initial state but may, or may not give any history. 
- ` TPHICgen` is the same a `TPythia6` 
- `TTherminator` uses status code 1 for decayed particles, and 0 for all other. 
- `TUHKMgen` sets status=0 for hydro-generted particles, and  1 particles produced in jets

Other interfaces used are 
- `TPythia6`, which uses status code 1 for final state, 11 for decayed, and 21 for initial state\
- `TPythia8` which uses 1 for final state, and 0 for everything else (including initial state)

This is what I could deduce from reading the code., 

So we should test for status values of 

- 3 for EG's that use the Pythia6 `HEPEVT` format (`THerwig`)
- 4 for EG's that adhere to the HepMC format 
- 21 for EG's that use the Pythia6 `PYJETS` format (`TPythia6`, `TPHICgen`, `TStarlight`, `THydJet`)
- The rest does not export the full event history and therefor we do not get the beam particles at all. 

Furthermore, we may need to introduce a check for status=11 when checking if a particle is from decay, though it seems not to be necessary. 

The odd one out is `TPythia8` which does not propagate the full status code. The best thing would be if `TPythia8` would use `Pythia::Particle::statushepMC` which however would require a change in ROOT. Optionally we could patch it up in `AliPythia8` or `AliTPythia8`.

Note, however, that in the old implementation, we discarded everything that had a status code larger than 1, which is potentially very bad, since that code strongly depends on the EG.

I don't know exactly what is going on with all this in the DPG, but surely this is something that should be addressed. 

Yours,

Christian

amorsch:
>I don't know exactly what is going on with all this in the DPG, but surely this is something that >should be addressed.
The point is that we need a IsPhysicalPrimary that works with existing MC productions. I conclude from the discussion that the code is not yet ready. Do you agree?

fprino:
Hi,

there is nothing going on in the DPG with this development.
The way the MC generators are interfaced to AliRoot is not under the DPG responsibility (to my knowledge).
Let me know if my understanding is not correct.

When in July we saw the e-maiils of Ivan and Nicolo' which were reporting the issues with the first implementation, we started to take an eye on this (I personally replied to Ivan suggesting some checks) because we thought it could potentially also affect the QA of the ongoing MC reconstructions and therefore it was relevant for some of the DPG activities.

Personally, I agree with the consideration of Andreas that we need an IsPhysicalPrimary that works with all the generators we use, and based on the message of Christian it seems we are not yet there.

amorsch:
if we replace
if (IsStable(m->GetPdgCode())) return false;
by 
Int_t mpdg =TMath::Abs(m->GetPdgCode());
if (IsStable(mpdg)  && mpdg!= 2212 &&  mpdg <1000000000) return false;

(to be tested)
we catch the cases when we loop back to the beam particles and do not 
depend on status codes at all.
what do you think ?

cholmcc:
I do not agree that it doesn't work - just take a look at Nicolò's test above to see that it does work for a given EG - despite that what I found from looking at the code, it shouldn't.  

So what I think we need to do is to run tests.  If someone could help me compile a list of simulation productions that would cover the relevant use cases, I could run a test.  Rather than looking at reconstructed tracks, I'd like at the simulation information only, and compare the distribution of number of primary particles per specie type.

BTW, does any of you know how to limit the number of events analysed on the Grid?  There's no reason to look at hundred of thousands of events when some tens of thousands will suffice. 

@amorsch I think you're proposed fix

```
Int_t mpdg =TMath::Abs(m->GetPdgCode());
if (IsStable(mpdg) && mpdg!= kProton && mpdg <1000000000) return false;
```

isn't correct. We can summarize the code by the following table

(Grand)Mother of p | is (anti)proton  | is nucleus       | otherwise
-------------------|------------------|------------------|------------------
is stable          | p may be primary | p not primary    | p may be primary 
is not stable      | p may be primary | p may be primary | p may be primary 

So when we track down the decay chain and we end at the beam particle (either a proton or nucleus), we see that beam particle is stable but either a proton or a nucleus and so the loop continues (and terminates since there's no more particles to track down) and the function returns `true` 

If we hit an intermittent proton or nucleus, we see that it is stable, but we do not flag the particle as secondary in that case (which we should).  Think of the (fictitious) chain 
```
p (beam) -> D0* -> pbar -> pi
```
Even though the pbar is stable, we'd not flag the pi as primary in that case. 

We could ask the (grand)mother if it has any mother, which would cure the above problem 
```
Int_t mpdg =TMath::Abs(m->GetPdgCode());
if ((mpdg == kProton || mpdg > 1000000000) && m->GetFirstMother() < 0) // Check for beam
  return true
if (IsStable(mpdg)) return false;
```
However, some EGs does not export the beam particles, and so the one may have the record 
```
pbar (not beam)  -> pi
K -> pi
deuteron -> n
```
Here, the pbar and deuteron would be wrongly considered as beam particles and the pion and neutron would be wrongly flagged as primary. 

@fprino I thought the DPG was in charge of running simulations too? And working to set-up models as external resources. Or maybe I misunderstood something.  If the DPG _is_ running simulations and working on the model interface, I guess the propagation of the status flags is of importance to you. 

Yours,

Christian

fprino:
Hi,

the DPG is indeed responsible for running the productions and QA them.
In this sense, yes, as DPG we are interested in this PhysicalPrimary development because it impacts on our activities. 
However, the setup and test of generators and model inderfaces in AliRoot is actually followed by PWG-MM, to my understanding.

I fully agree that we have to run tests with different event generators to make sure the new code works in all cases.
The most commonly used event generators for "general purpose" productions are PYTHIA6 and PYTHIA8, EPOS, HIJING, DPMJET.
But then PWGs have their specific generators: UD has a bunch of them (STARLIGHT...), DQ uses a custom generator for J/psi kinematics (if I am not wrong), CF is trying to revive AMPT...
So, it may be that we have quite a number of cases to cover.

For limiting the number of events, if you run private jobs on alien you can do it via the 3rd argument of AliAnalysisManager::StartAnalysis:
   Long64_t            StartAnalysis(const char *type, TTree * const tree, Long64_t nentries=1234567890, Long64_t firstentry=0);

Another question: shouldn't move this discussion to a Jira ticket, where we could add in CC the PWG-MM conveners etc?
I know that the PR is more specific to discuss code developments, but if you plan to proceed with testing generators and report results, it seems to me that Jira is the place in which the other Alice members would expect to find the discussion.

Cheers, Francesco

amorsch:
Christian, 
I applied my proposed changes and it works at least with Pythia6. No difference between old and new for 1k events. I'll continue the tests with other generators and let you know. 
Concerning your counter example: if the pion comes from a secondary collision simulated be the event generator we would consider it as primary. It comes from the O(fm) collision regions.

Andreas

chiarazampolli:
Hello,

I confirm what Francesco said: "the setup and test of generators and model inderfaces in AliRoot is actually followed by PWG-MM", while DPG runs and QA'es productions. We also provide tools for the analyzers, so we got interested in this issue. 

For MM, I am adding @qgp here, but I agree that a JIRA ticket is better, also because any change in aliroot that affects significantly productions and analyses should be tracked there, as discussed at the CB last Monday. 

Chiara

chiarazampolli:
Hello,

JIRA created:

https://alice.its.cern.ch/jira/browse/ALIROOT-7437

I propose to continue there, or keep it anyway to have track of this for the next tag.

Chiara

cholmcc:
@amorsch 

I'm not sure what you mean by 

> ... if the pion comes from a secondary collision simulated be the event generator ...

The event generators does not simulate secondary collisions (e.g., interactions with material) as far as I know.  The transport simulation does, but stores the process code in TParticle::fUniqueID - which is why we check that. 

In my cooked up example 
```
pbar (not beam)  -> pi
```
I was considering a decay - which of course have more than a gazillion years of mean life time, so not really that relevant. Similar goes for at least deuteron and alphas. 

I tried your fix on Hijing which also works.  Note, the initial check on the status code _must_ be removed (I think it's probably superfluous for EGs)

@fprino 

The new implementation (minus Andreas' fix) is good for all EGs - _except_ in cases where the full event history is exported by the EG. However, _most_ EGs does not export the full event history, as outlined above.  DQ - as far as I know - cares not about whether a particle is _primary_ or not, but rather whether it is _prompt_ - which probably needs to be defined also.  The only EGs that I know of that export full histories are Pythia (6 and 8), HERWIG, HydJet, and Therminator.  All others cut the event history - including EPOS, DPMJet, and HIJING. 

Concerning 
> `Long64_t StartAnalysis(const char *type, TTree * const tree, Long64_t nentries=1234567890, Long64_t firstentry=0);`

If you check the code, you will see that the number is propagated verbatim to the worker, which means it is the number of events per worker - a number one can not easily predict. 

Do you have some "typical" examples of productions I can test the code on?  Just the production identifier is fine.  Thanks.

Yours,

Christian

amorsch:
> ... if the pion comes from a secondary collision simulated be the event generator ...
https://arxiv.org/pdf/hep-ph/9911232.pdf
page 31ff

or run tests on dpmjet output
A.

ktf:
Is this still relevant? If not can it be closed?

dberzano:
There is an ongoing discussion on JIRA but to my understanding it never converged. I am closing this for now, but it can be reopened at any time when/if such changes are agreed upon by the authors and the Physics Coordination.

PR #348, opened 2017-08-09
Sync. macros whith version used to produce current recoparam

shahor02:
The TPC/Calib/RecoParams of 2017 have setting to use the Lumi info from the SP 
correction map (if available), synced macros producing recoparams to reflect current status


shahor02:
+1

PR #351, opened 2017-08-10
K0L stable in decaytable and protection against changes in AliDecayerPythia

amorsch:
see also 
https://alice.its.cern.ch/jira/browse/ALIROOT-6940


chiarazampolli:
+1

FriederikeBock:
@amorsch how would we then for the cocktail allow this decay, as we need it for the corrections for the pi0's (which was the reason @laltenka had put it like this)

chiarazampolli:
@FriederikeBock , so this was intentional?

FriederikeBock:
@chiarazampolli, let's say semi-intentional. We need to be able to let Pythia decay those particles if asked to do so and I guess @laltenka saw this as the easiest way to do this, without being aware of the implications this has for other purposes. Sorry for causing so many troubles.

chiarazampolli:
Ok, I see. Note that I approved this PR, and it was merged. 

Chiara

amorsch:
@FriederikeBock 
you can use 
AliDecayerPythia::DecayLongLivedParticles()
I suppose you take into account the decay vertex in some way. Most of the K0L
get absorbed before they actually decay I would guess.
Andreas

PR #357, opened 2017-08-11
ALIROOT-7333 Do not use NystromRK4 stepper, it is not working correctly

pzhristov:
Use the default Geant4 stepper in the tests: NystromRK4 has problems as discussed in 

 https://alice.its.cern.ch/jira/browse/ALIROOT-7333


pzhristov:
+1

PR #359, opened 2017-08-11
Correct x-position of ZEM support

ecalvovi:
Support plates of ZEM should be centered w.r.t beam pipe in x-direction.


pzhristov:
+1

PR #361, opened 2017-08-13
 Check for duplicates, fill muliple labels, use all tracks in case of embedding

shahor02:
1) when filling nCl-per-track tree check for duplicate labels in the same cluster 
2) fill additional array nclMPerTrack for every MC label, accumulating there total non-negative labels count for clusters contributed by this label. I.e. if all N clusters related to given label were made of this label only, the nclMPerTrack[label] will be equal to N, if all clusters had exactly 2 non-negative labels contributing, it would be 2N etc. Therefore, statistically each nclPerTrack[i] should have a weight float(nclPerTrack[i])/nclMPerTrack[i]<1 to add up to nclTot-nclOrphan


shahor02:
+1

PR #364, opened 2017-08-15
bugfix in the task configuration of the analysis manager

mazimm:
define __R_ADDTASK__ if first use is in a scoped block of the task configuration


dberzano:
+1

PR #370, opened 2017-08-21
Restore the functionality of some tests

pzhristov:
Use ALIROOT_OCDB_ROOT to find the default OCDB. Fix the loading of the magnetic field. Load libpythia to avoid crashes when the header is stored in galice.root.


pzhristov:
+1

PR #374, opened 2017-08-24
fixing include path for Pythia8 headers

zwound40:
Following the discussion on the alice-analysis-operations@cern.ch mailing list here is the fix to use Pythia8 in aliroot without having to add the Pythia8 folder to ROOT's include path


dberzano:
+1

I agree with the pull request, it makes sense. This makes the following superfluous:

```
gSystem->AddIncludePath("-I$ALICE_ROOT/include/Pythia8")
```

This means that this PR does not break any current use case.

PR #375, opened 2017-08-26
Anti-particles of heavy vector mesons (Geant4)

pzhristov:
For some reason it is not enough to define only particles, Geant4 needs separate definition of antiparticles. The problem was detected in test/genkine.


pzhristov:
+1

PR #384, opened 2017-08-30
Xi(1820) states added

mpuccio:
Please wait for the feedback of resonance people before merging!

### DIFF
--- a/STEER/STEER/AliMC.cxx
+++ b/STEER/STEER/AliMC.cxx
@@ -330,6 +330,31 @@ void  AliMC::AddParticles()
   // f2(1270) (PDG: width = 185 MeV)
   TVirtualMC::GetMC()->DefineParticle(225, "f2_1270", kPTNeutron, 1.275 , 0.0, 3.558e-24,"Hadron", 0.185, 0, 0, 0, 0, 0, 0, 0, 0, kTRUE);
 
+  // Xi_0(1820)
+  TVirtualMC::GetMC()->DefineParticle( 123324,"Xi_0_1820",    kPTNeutron,1.8234,0.0,2.742550e-24,"Hadron",0.24, 3, -1, 0, 1,  1, 0, 0,  1, kTRUE);
+  TVirtualMC::GetMC()->DefineParticle(-123324,"Xi_0_Bar_1820",kPTNeutron,1.8234,0.0,2.742550e-24,"Hadron",0.24, 3, -1, 0, 1, -1, 0, 0, -1, kTRUE);
+
+  int xi_0_1820_mode[6][3] = {{0}};
+  float xi_0_1820_ratio[6] = {100.f,0.f,0.f,0.f,0.f,0.f};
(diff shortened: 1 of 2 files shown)


dberzano:
@amorsch can you please have a look? Thanks!

dberzano:
@amorsch ping

amorsch:
The code itself  is ok, but I wonder what is the exact purpose.
I think there is no need to transport the Xi(1820) if it is produced and decayed by the event generator. So the  TVirtualMC::GetMC()->DefineParticle(...) calls are not really needed.
There are many other particles which are not known to Geant3/4 since they are handled by the event generators. 
pdgDB->AddParticle(...
is a different story since the data base entry is needed when passing the particle to the AliRoot stack.
Cheers
Andreas

mpuccio:
Hi Andreas,
we will inject the Xi(1820) with AliGenBox. I am not an expert, does it handle the decay of particles?
I always assumed it does not since also other resonances not produced by default by the EGs are handled by the transport code.

amorsch:
It is correct that AliGenBox does not perform decays. I recommend to use AliGenParam instead. It performs the decays and has more flexibility.
Andreas

mpuccio:
As far as I see AliGenParam uses the PYTHIA decayer (sorry if I missed something).
Should we then add the states to the PYTHIA decay table? If so, just for my education, what is the advantage wrt use GEANT to decay these states?

amorsch:
Yes, AliGenParam can use the Pythia decayer. The advantages are that we do not have to maintain yet another decay table and that primary particles are produced by the primary particle generator and not by the transport code. There might be also cases where the simple phase-space decays performed by Geant are not precise enough.

bdoenigu:
Dear Andreas,
this is a general strategic problem. We already have the f states and the resonant exotica (+hypernuclei and weakly decaying dibaryons) in AliMC, so we either have to have one place to put all at the "unknown" things or a good reason why we put one here and the other there. I remember that Federico Carminati and Peter Hristov were in favour to add those states to AliMC. I understand your reservation to maintain two classes doing similar things but there was probably a reason why we did that in the first place, which I can not remember...
Cheers,
Benjamin

amorsch:
Dear Benjamin,
for weakly decaying particles (long life-time) a decay is only one of the possible processes during transport and it has to be treated by the transport code. Also in this case it is possible to use the external decayer which I recommend. In all other cases, despite of what was done in the past, it is more reasonable to have the decays as part of the primary event  generation for the simple reason that the decay products *are*  primary particles. What are the arguments against this strategy ?
Regards
Andreas

mpuccio:
Hi,
I can add the particles to any decay tables, just give me a pointer to them because I don't know where to put the modifications.
Cheers,
Max

amorsch:
Hi Max, you have to add them to 

`data/decaytable.dat`

which is a patch to the Pythia6 decay table.
There are already many examples, so I think it's self-explanatory.
Cheers
Andreas

PR #385, opened 2017-08-31
New SDD macro to read CDH Block Attributes

mario6829:
A new macro was created to easily read the content of the SDD CDH Block Attributes in the data (which will contain the firmware version number).


fprino:
+1

PR #387, opened 2017-08-31
Update runMap to include 2017

miranov25:
Needed for the QA of the 2017 data


dberzano:
+1 but I am wondering whether we can do it in a way more appropriate way for the future (asking @mkrzewic to this respect). We have this information duplicated in way too many places and we unfortunately need to keep it in sync.

miranov25:
Hallo.

I also do not like the way how we get the run ranges for the QA now. Unfortunately I'm not aware of DB with shell  interface where we can easily access  information which we need.
I prefer to do not rely on the alien. Maybe we can rely on cvmfs?

Marian

mkrzewic:
@miranov25 fully agreed, what we have in there is a hack from the old days. If this information exists on cvmfs we can parse it probably even without changing the bash API.

dberzano:
There's the XML file with the run mappings, both on AliEn and CVMFS, but the collision type is missing. In principle one should directly use the OCDB, but you need AliRoot for that. Maybe the best thing would be to have this file generated (and then committed) straight out of the OCDB information.

@mkrzewic whenever I hear "bash" and "API" in the same sentence I admit I shiver a bit :-)

mkrzewic:
@dberzano how about "ASI" application scripting interface? ;-)

dberzano:
You're really pulling my leg here, lol :-)

PR #389, opened 2017-09-01
rename Momentum method to GetMomentum to avoid compilation warning du…

gconesab:
…e to method in AliVParticle with same name but different parameters


gconesab:
Hello, can someone approve this simple commit that fixes a compilation warning?
Thanks

dberzano:
+1

PR #392, opened 2017-09-04
Include ROOT headers as -isystem

davidrohr:
This pull request includes all root headers in HLT and in STEER/STEER as -isystem, to avoid fake warnings coming from within the headers during AliRoot Compilation.
I would suggest to apply this to all AliRoot libraries.
I have added STEER, such that this will not be merged automatically.

### DIFF
--- a/HLT/BASE/CMakeLists.txt
+++ b/HLT/BASE/CMakeLists.txt
@@ -28,13 +28,11 @@ endif()
 set(MODULE HLTbase)
 
 # Module include folder
-include_directories(${ROOT_INCLUDE_DIR}
-                              ${AliRoot_SOURCE_DIR}/HLT/BASE
-                   )
+include_directories(${AliRoot_SOURCE_DIR}/HLT/BASE)
 
 # Additional include folders in alphabetical order except ROOT
-include_directories(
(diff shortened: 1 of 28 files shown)


dberzano:
+test

dberzano:
OK let's leave it open to add it everywhere. @davidrohr can you make it "editable by maintainers"?

davidrohr:
@dberzano : sorry, didn't follow this up or some time while I was on conference nd then vacation.
What do you mean by "editable by maintainers"? There is a box n the pull request page, but that is already checked.

i am wondering what is the failure reported by alibuild? It says 0 tests failed, and the link to the full log doesn't work for me.

pzhristov:
Hi @dberzano, do you mean that we have to apply the same changes in all our CMakeLists.txt? I agree, but we can do it in a separate pull request. Lets merge this one. If nobody has objections, I will do it before lunch.
Cheers,
Peter

davidrohr:
Hi @pzhristov, yes, the same change would need to be applied to all CMakeList.
I first wanted to do it in a global way, but due to the way the root headers are included, this is not possible.

pzhristov:
+1

PR #393, opened 2017-09-04
Enable coalescence afterburner on LEGO trains (@eserradi)

mpuccio:
This macro will make possible running the coalescence afterburner coded by @eserradi on MC LEGO trains.
For the records: the results obtained with the coalescence afterburner were presented by @eserradi and @akalweit at the PF on the 25th of August.


dberzano:
+1 this macro is harmless and not used by default, approving

PR #394, opened 2017-09-05
Fix unhandled case AliDecayerEvtGen

aaniin:
A few decay cases, see the full enum in AliDecayer.h, were not handled. Replaced explicit listing of decays by a default

### DIFF
--- a/TEvtGen/TEvtGen/AliDecayerEvtGen.cxx
+++ b/TEvtGen/TEvtGen/AliDecayerEvtGen.cxx
@@ -204,8 +204,7 @@ void AliDecayerEvtGen::ForceDecay()
   // Set the decay mode to decay particles: for each case is read a 
   // different decay table. case kAll read the default decay table only   
   //
-  Decay_t decay = fDecay;
-  switch(decay)
+  switch(fDecay)
     {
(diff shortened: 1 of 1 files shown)


dberzano:
Hi @aaniin, while the default case makes more sense to the compiler, the explicit list of cases not covered can have some sense when someone decides to implement one or more of them (right?)

Does this fix some warning? In case it does, I suggest you at least add a comment with the list of the "default" cases for convenience.

aaniin:
Dear @laltenka ,

Can you give us a some insight on what the purpose of a a new decay type like kAllEM as introduced by you in https://github.com/alisw/AliRoot/commit/9324c24918e6aee5028dafaeca43b0af0e232427 without a decay table would be?
Do you ever intend to implement a decay table for this decay type?

Cheers,
Hans

laltenka:
Hi,

the intention was to combine kGammaEM, kElectron if I recall correctly. Maybe @FriederikeBock can elaborate. However it's true, without decay table this is not of much use. It probably was intended to implement it but somehow got lost.

Cheers,
Lucas

ktf:
@aaniin @laltenka @qgp can this be closed?

dberzano:
@aaniin's no longer with us - don't be scared, he's in a better place :-) But yes, the question still stands. What do we do with this PR?

laltenka:
Hi, from my side I think this PR can be closed.

PR #400, opened 2017-09-08
ATO-83 & ATO-46 MC to anchor comparison

miranov25:
* ATO-46 AliExternalInfo:GetMCPassGuess returns now also the anchor period name
  * fixed bug in anchor pass guessing: in case AliRoot, AliPhysics and production name matched, the wrong pass guess was provided
* ATO-83 error handling in case plotting fails
  * updated Doxygen documentation
* ATO-83 MC to anchor comparison
  * Behaviour of TStatToolkit::MakeMultGraph in case of empty plot for a variable can now be swichted between these options:
    * Demand a plot for every variable or otherwise return empty plot with error msg
    * Accept missing plots for some variables as long as not all are missing, if all are missing return empty plot with error msg


miranov25:
+test

miranov25:
For a record. 
* In addition to the standard AliRoot test also non standard AliPhysics QA test was running
successfully.
* Pull request for updated test in AliPhysics is under preparation.
Test:
```
( source $AliPhysics_SRC/PWGPP/QA/detectorQAscripts/test/unitTest.sh;  setupUnitTestMI; testQAT0; testQATPC;) | tee unitTest.log
```

miranov25:
+1

PR #401, opened 2017-09-09
AT0-83 - Fix for doxygen documentations for TStatToolkit

miranov25:
* using \namespace instead of \class doxygen directive
* small modification in documentations
* modifying curly brackets in some functions
  * once full documntation provided consistent usage of the brackets should be requested


miranov25:
+1

PR #403, opened 2017-09-10
Introduction of AliESDcascade::RefitCascade method

ddobrigk:
This change introduces a method in AliESDcascade to improve the precision of the cascade decay position estimate by using an error-weighted average for that estimate, as discussed initially [here](https://indico.cern.ch/event/658280/contributions/2685555/attachments/1505439/2345751/DDChinellato-VertexerImprovements-7.pdf). The final implementation employs a covariance matrix calculation that's analogous to what is already done in the AliESDv0::Refit method and borrows from the same function used there for single track covariance estimates (see "GetWeight"). The ::RefitCascade function, as contained in this pull request, is also extensively commented for clarity of purpose and understandability. 

As a side product, the RefitCascade method will also produce a chi2 value for the cascade decay vertex and a cascade position covariance matrix - again analogous to what takes place within AliESDv0::Refit. 

This change will not affect any data members (and hence no ClassDef incrementation was done) and will not affect any current use case, given that unless ::RefitCascade is invoked nothing different will happen wrt current operation. 

Update: Corresponding JIRA Ticket:
https://alice.its.cern.ch/jira/browse/ALIROOT-7485

### DIFF
--- a/STEER/ESD/AliESDcascade.cxx
+++ b/STEER/ESD/AliESDcascade.cxx
@@ -16,18 +16,20 @@
 /* $Id$ */
 
 //-------------------------------------------------------------------------
-//               Implementation of the cascade vertex class
+//             Implementation of the cascade vertex class
 //              This is part of the Event Summary Data 
 //              which contains the result of the reconstruction
 //              and is the main set of classes for analaysis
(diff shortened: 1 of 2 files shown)


iouribelikov:
+1
I am fine with these changes !

dberzano:
+1 thanks @iouribelikov!

PR #405, opened 2017-09-12
PWGPP-312, ATO-83, ATO-46: Update  TStatToolkit::MakeMultGraph

miranov25:
### ATO-418: Updated styles for arrays attributes ￼…
* check ranges of style array
  * handle out of range request. Returning invalid value instead of crash
  * interpret separated string as array of attributes
* Updated/improved doxygen documentation
### ATO-83 & ATO-46: Use AliDrawStyle to specify graph attributes
  * keeping back compatibility (old - string with arrays of values, of style ID)
### 	PWGPP-312: MakeMultiGraph with cut array
* see Doxygen documenation in the  TStatToolkit::MakeMultGraph
  * expr to drow and cuts  can be  array
  * before only expression could be array
```
TMultiGraph * TStatToolkit::MakeMultGraph(TTree * tree, const char *groupName, const char* expr, const char * cut, const char * markers, const char *colors, Bool_t drawSparse, Float_t msize, Float_t sigmaRange, TLegend * legend, Bool_t comp){
///  \param expr       - string <variablearray>:<tagID>:<variablearrayError>
///                      <variablearray>="var0; ...; varN
///                      <variablearrayError>="err0; ...;errN"
///  \param cut        - string "selection", resp. cut array
///                    - <variablearray>="defaultCut;cut0;cut1;...; cutN"
```
* markers and colors array provide using AliDrawStyle interface
  * either style name (ID) or array of codes

### DIFF
--- a/STAT/AliDrawStyle.cxx
+++ b/STAT/AliDrawStyle.cxx
@@ -15,15 +15,15 @@
 
 
 ///  ### AliDrawStyle - Class to access drawing styles
-///  * Several drawing styles can be regeistered and used in the same moment
-///    * Styles are identified using strings as identifiets
+///  * Several drawing styles can be registered and used in the same moment
+///    * Styles are identified using strings as identifiers
 ///      * TStyle
 ///      * MarkerStyle[]  AliDrawStyle::GetMarkerStyle(const char *style, Int_t index);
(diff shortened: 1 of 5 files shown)


miranov25:
+1

PR #407, opened 2017-09-13
Avoid issues with the latest cling

pzhristov:
The forward declaration of AliRawDataArrayV2 is correct, but the latest cling doesn't accept it. Until we get a fix in Root, I propose to include the header file.


pzhristov:
+1

PR #409, opened 2017-09-16
fix copying of aodheader

loizides:
This fix is needed in order to be able to skim AOD files, without the fix the header info would not be copied.
Note I believe someone has to update the code as not all branches of the AOD are treated (like TZERO, ZDC, and ADA plus more are missing).

This fix is rather urgent.


pzhristov:
Hi @loizides,
Even if this is a static data member, shouldn't we increment the ClassDef?
Cheers,
Peter

dberzano:
Any news on this PR? Otherwise I'll close it and we can come back to it later.

pzhristov:
I think we can accept it.

dberzano:
@pzhristov I think by the way that ClassDef should not be updated as the commit concerns a `//!` member. If you are OK with that, please `+1`!

pzhristov:
+1

PR #411, opened 2017-09-19
TPC eloss modelling in StepManager instead of G3/G4 (P.Christiansen)

shahor02:
TPC stepmanager will simulate itself the eloss instead of using Geant, details in
https://alice.its.cern.ch/jira/browse/ATO-425


shahor02:
+1

PR #413, opened 2017-09-20
changes in the fluka vmc scipts to write the error output from lun 15 into fluka.err

amorsch:
needed for new version fluk2011_2c


amorsch:
Somehow the pull-request for AliStack (see ticket AliROOT-7477 got bundled with the one for FLUKA.
FLUKA has no dependencies with the production software. So this is not a problem.

pzhristov:
+1

PR #415, opened 2017-09-26
Ali option parser fix

### LINKED ISSUE #189: Undefined symbols in HLT libraries
If AliRoot is build without ZeroMQ, the `AliOptionParser` code is not compiled into `libAliHLTUtil`. Still the header file is included in some components and even used as a base class, e.g. `AliHLTTPCClusterStatComponent`. The corresponding library can not be loaded or linked due to undefined symbol.

@mkrzewic: `AliOptionParser` should be in a separate file and independent of availability of ZeroMQ.

mkrzewic:
this fixes #189


mkrzewic:
It seems this PR creates a few additional problems for the future - i'm closing it for now, will resubmit when issues are fixed

matthiasrichter:
HI @mkrzewic , is it possible to make a smaller set of changes?

mkrzewic:
@matthiasrichter it is possible, but not practical. I'll see if I can squash some of it for a smaller number of commits.

matthiasrichter:
ok, but I though more of eliminating the reason for stopping this PR.

mkrzewic:
that would be the new class in STEER I imagine. This is depended upon by other commits, so it has to stay in there - and be +1'd by somebody. I'll separate this out in a single commit so that it is clear what is going on there

mkrzewic:
@matthiasrichter BTW i noticed we are commenting on a PR which I closed and replaced with #419

PR #416, opened 2017-09-26
introduce string attributes in NanoAODs

mazimm:
allow NanoAOD header files to also save string variables
remove some compiler warnings from the nano AOD framework

### DIFF
--- a/STEER/AOD/AliNanoAODHeader.cxx
+++ b/STEER/AOD/AliNanoAODHeader.cxx
@@ -12,6 +12,7 @@ AliNanoAODHeader::AliNanoAODHeader():
   fCentrCL0(-1),
   fCentrCL1(-1),
   fMagField(-1),
+  fOfflineTrigger(-1),
   fRunNumber(-1)
 {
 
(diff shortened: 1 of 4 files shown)


mazimm:
+test

jgrosseo:
+1

jgrosseo:
I approved this. However, we should discuss if saving a string is the best way for a nanoAOD where we want to reduce the data volume. This trigger string list is large and quite redundant. We could consider the following solution:
- When filtering the nanoAOD, we allow to filter on a set of trigger strings
- These trigger strings (and only that ones) are mapped into an integer, and that integer is saved as a custom variable

mazimm:
The point here was to have the structure of the NanoAODs as similar to the AODs as possible. With this implementation the user has to change nothing. I am up to discuss alternative solutions.

jgrosseo:
Yes, I think it makes sense to go for a different approach here in the mid-term. Could you ask Ruediger what he does exactly with the information?

mazimm:
This is only for a control histogram. So there your approach works well. The more interesting question is if there are other use cases which do not have a finite list of strings which are used. 
This should be a general development which applies to all string functions.

jgrosseo:
Isn't this the only string one we have?

mazimm:
Oh, I didn't notice. You are right, it is the only one. I can change it to the way you suggested if I get a full list of all possible Strings which can fill this one. 
Do we ever expect more string methods?

mazimm:
and then we define another Map in the AliNanoAODHeader which is used to look up the correct bit. In this case we can save the information in the normal double vector which is used for the other variables. That's very efficient.

rhaake:
Hi,
concerning this control histogram, I do not even use it. It is contained within the EMCal jet framework.
So from my point of view, it could also be empty. However, it would be good if a general EMCAL jet framework task would be supported for NanoAODs out-of-the-box.

Maybe a bit philosophical, but in my opinion, it is not a good idea to force the user to define the trigger list on his/her own. Those lists could differ from dataset to dataset and it's not so trivial to get them (and they even depends on the runlist?). In short, I think it could prevent people from getting this to work properly or even to prevent them to general NanoAODs at all.

What one could also consider is to make this string optional. This would be fine for my analysis but in general also compatible to EMCAL jet framework tasks if they need the information.

PR #417, opened 2017-09-26
Fixes needed for compilation with Xcode 9

mfasDa:


mfasDa:
clang from Xcode9 finds an ambiguity between two string constructors here which has to be manually resolved.

pzhristov:
+1

PR #421, opened 2017-09-28
 Store mother particle and decay history in kine tree and two decay tables

mbroz84:
Analysis of J/psi yield enhancement in peripheral events wants to use Starlight event in cocktail with Hijing. Therefore we need mother particle and history of decay to identify the UPC electrons. EvtGen decay tables are for J/psi to lepton decays with QED radiative corrections via Photos package. Whole thing is to re-decay Starlight J/psi in EvtGen with radiative corrections and put it to Hijing event.

### DIFF
--- a/STARLIGHT/AliStarLight/AliGenSLEvtGen.cxx
+++ b/STARLIGHT/AliStarLight/AliGenSLEvtGen.cxx
@@ -103,7 +103,6 @@ void AliGenSLEvtGen::Generate()
   TLorentzVector momGen;
   TLorentzVector momTot;
   momTot.SetXYZM(0.,0.,0.,0.);
-  
   static TClonesArray *decayProducts;
   if(!decayProducts) decayProducts = new TClonesArray("TParticle",1000);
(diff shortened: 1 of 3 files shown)


rbertens:
+test looks good, @qgp, could you also check?

rbertens:
@ktf @dberzano  @mbroz84 apologies for my ignorance, but could one of you elucidate what the sw/BUILD/AliRoot-guntest does (and also if it's failure is in this case problematic)? thanks,
Redmer

mbroz84:
There is one Fatal in the log which comes from PIDresponse set-up in CheckESD. It seems to have nothing to do with my generator update, generator even didn't run at all...

ktf:
AliRoot-guntest is the usual simple "gun" test from https://github.com/alisw/AliRoot/tree/master/test/gun . The error seems to be transient. @dberzano any idea?

dberzano:
Yes, indeed, this happened once already ant it's unrelated to the guntest itself. The error was transient.

rbertens:
+1 ok, then let's no longer delay this PR

PR #422, opened 2017-09-28
Account for beam A/Z in LHC B-field initialization

shahor02:
Pass A/Z info to field creation to properly rescale rigidity for different ion types


shahor02:
+1

PR #426, opened 2017-10-03
make storage of Fired Trigger Particles more efficient in the NanoAODs

mazimm:
change the internal storage of the fired trigger particles to a coded int per event instead of full strings. This requires less disk space.

### DIFF
--- a/STEER/AOD/AliNanoAODHeader.cxx
+++ b/STEER/AOD/AliNanoAODHeader.cxx
@@ -33,9 +33,10 @@ AliNanoAODHeader::AliNanoAODHeader(Int_t size):
   fRunNumber(-1)
 {
   AllocateInternalStorage(size, 0);
+
 }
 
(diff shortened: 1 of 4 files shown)


dberzano:
+test Is the change transparent for users? Is there a JIRA (albeit not needed) where this was discussed?

mazimm:
This was discussed here:
https://github.com/alisw/AliRoot/pull/416#issuecomment-332535048

It has only an effect on Nano AOD generations which include the information about Fired Trigger Classes. This is not the case up to now. The first use will be introduced in AliPhysics as soon as this is available in the next tag.

dberzano:
So basically if we merge this, it won't break current AliPhysics?

jgrosseo:
+1

mazimm:
No it won't. 
I assumed that your test which was executed here would also test the compilation with AliRoot. Is this not included yet?

jgrosseo:
Yes, doesn't break anything as the original feature was only added last week (and is not used yet)

PR #427, opened 2017-10-04
Cap PARfiles build cores on the Grid

dberzano:


dberzano:
@costing fyi (we'll do the same for AliPhysics).

PR #428, opened 2017-10-04
Remove dependency on Eve

ktf:


ktf:
+test

ktf:
@pzhristov @davidroh I suspect this is the culprit for the library loading reported in last CB. I cannot find any JIRA issue about this.

dberzano:
Here's the [relevant comment from JIRA](https://alice.its.cern.ch/jira/browse/PWGPP-364?focusedCommentId=204287&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-204287) for the record.

The error is the following:

```
dlopen error: libGLU.so.1: cannot open shared object file: No such file or directory
Load Error: Failed to load Dynamic link library /cvmfs/alice.cern.ch/x86_64-2.6-gnu-4.1.2/Packages/ROOT/v5-34-30-alice7-14/lib/libEve
```

and it's due to the lack of OpenGL libraries on that specific worker nodes. To recap: in theory, since the event visualization part is not used on Grid mass processing, this library should not depend on Eve (which depends on OpenGL in turn). The compilation results are positive... so this PR in my opinion is OK, but it would be better to have some feedback from the experts before proceeding.

pzhristov:
+1

PR #429, opened 2017-10-04
Modify TriggerSTU,TriggerTRU to not run the trigger simulation if the thresholds…

moliver813:
… are missing.

Error messages given during TriggerElectronics creation.

### DIFF
--- a/EMCAL/EMCALsim/AliEMCALTriggerElectronics.cxx
+++ b/EMCAL/EMCALsim/AliEMCALTriggerElectronics.cxx
@@ -109,6 +109,25 @@ fGeometry(0)
     }
   } else fSTUDCAL = 0;
 
+  // Checking if the L1 thresholds are missing
+  for (int ithr = 0; ithr < 2; ithr++) {
+    if (fSTU->GetThreshold(kL1GammaHigh + ithr) == 0) {
+      AliError(Form("EMCAL DCS STU has 0 threshold for EGA %d.  This trigger will not be simulated.  Check OCDB!!!",ithr));
+    }
+    if (fSTU->GetThreshold(kL1JetHigh + ithr) == 0) {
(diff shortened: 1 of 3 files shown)


mfasDa:
+test

In my opinion OK, this change will lead to no patches when having unexpected behavior (0-threshold in OCDB), which should be catastrophic enough to make people aware of problems.

mfasDa:
+1

As the PR was open a sufficient time and the change looks OK I approve it. Build errors seem to be unrelated to the commit. One remark: To better keep track of PR it would be good to add a JIRA ticket Number if available (linked to ticket - here it is) at the beginning of the PR.

@moliver813 @gconesab Do we need to back port to versions older than master (i.e. patches branches)?

gconesab:
Hello @moliver813 and @mfasDa 

I though we said we would call an AliFatal, if I am not wrong, this commit will create tons of messages with AliError not solving the issue if the logs are not checked and keep reasonable sizes.

Regards ... Gustavo

mfasDa:
Hi @gconesab ,

Indeed we agreed on AliFatal, thanks for spotting! @moliver813 , could you please move the AliErrors to AliFatals?

Cheers

Markus

moliver813:
Yeah, I can make that change.  I also tried to make sure my errors only occurred at the beginning of a simulation, but I apparently I didn't do that for the L0 error messages.  What if only a single TRU is missing a threshold?  Should I still call an AliFatal?  I assume yes, since that would produce either invalid results, or cause the L0 bug again.

mfasDa:
Yes, it should crash with an AliFatal rather then give invalid results.

gconesab:
Hello, @moliver813 
I have not seen the implementation of the fatal?
Thanks ... Gustavo

PR #431, opened 2017-10-09
Add missing include <stdexcept>

mkrzewic:
reported by Barth: std::runtime_error needs #include (at least on SLC6 with gcc 4.9.1)


dberzano:
Please do not approve yet, I am making a change in permissions (adding @mkrzewic) and I'd like to test them :-)

PR #434, opened 2017-10-11
Remove duplicated AliOptionParser entry

raymondEhlers:
When ROOT 6 is launched, there is a warning:

> Warning in <TInterpreter::ReadRootmapFile>: class  AliOptionParser found in libAliHLTUtil  is already in libAliHLTGlobal

I suspect that the class was not removed from the link def when it was relocated. This should resolve that warning.

@mkrzewic - Would you prefer this pull request here or on the HLT fork?


mkrzewic:
Hi this was already fixed in the HLT fork: 8e45ec9b32a69949bbceef393468b2e90856c10c
i'll cherry-pick the commit.

mkrzewic:
this can be closed, fixed with #435

raymondEhlers:
Ah, of course! I didn't think to check if it was fixed in the HLT fork - sorry about that. Thanks for the quick fix!

PR #436, opened 2017-10-12
TOF trigger map in raw data: fixed LTM indexing

noferini:
- LTM indexing moved fro AliTOFTrigger to AliTOFGeomety (TOFbase dir)
- fixing for LTM indexing in raw data (reconstruction)


noferini:
It seems we need a fix also for mapping in MC.
Please wait a bit for an additional fix.
Thanks
Francesco

chiarazampolli:
+1

PR #441, opened 2017-10-13
Corrected list and order of libraries

pzhristov:
Corrected list and order of libraries: libDPMJET contains DPMJET, PhoJet and Pythia6115


pzhristov:
+1

PR #443, opened 2017-10-13
refinement on MC maxipad mapping

noferini:
- refinement of TOF  Maxipad indexing in MC
- add macro to draw the indexes on the supermodule


chiarazampolli:
As this was discussed in Jira, i am adding +1.

chiarazampolli:
+1

PR #444, opened 2017-10-16
Fix issue with ROOT6 when merging with the alien plugin

dstocco:
The merging of grid outputs is broken with ROOT6 due to a bug in an interpreted line of the AliAnalysisAlien.
The bug is fixed with this commit.


zconesa:
+1

PR #445, opened 2017-10-16
Revert "Fix issue with ROOT6 when merging with the alien plugin"

dstocco:
This reverts commit fbb64a6bab6d73aa1e3397c451b2fc8911cff575.

The fix causes issue with ROOT5: reverting for the time being.


zconesa:
+1

PR #451, opened 2017-10-18
fix for TOF maxipad channels swapped in the TRMs

noferini:
This fix is requested because we found some TRM links swapped.


chiarazampolli:
+1

PR #452, opened 2017-10-18
Adding possibility to set a max nb of mixing events

dcaffarr:


jgrosseo:
Hi @dcaffarr 

Thanks for this improvement. I have two comments:
- could we make -1 the default in case the fMixDepth is not set? So also change the .h file to that default
- I think the code produces a memory leak because the ownership of "trk" is with the event pool. So in case you don't store the event, you should delete it. However, wouldn't it be better to store this event, and remove the first one in the list. LIke we do it when the total number of tracks is exceeded? In this way the pool always has the last X events, and not the first X events which were analyzed.
See fTargetTrackDepth, maybe you can just set removeFirstEvent in case the number of events is exceeded and use the same functionality...

jgrosseo:
+1

PR #453, opened 2017-10-18
using light-yields in AD MC simulation

hcab14:
- photo-cathode efficiency times light-yield is taken into account in the MC simulation, analogously as for VZERO
- the old behavior can be restored by setting AD/Calib/LightYields entries to 1
- nPhotonsPerMIP was adjusted
- code cleanup


ecalvovi:
+1

PR #456, opened 2017-10-19
Fix handling missing X11 libraries (on macos)

mkrzewic:
when missing, throw a warning and don't build ZMQhistViewer

this should fix ALIROOT-7555


dberzano:
+1 and many thanks!

PR #457, opened 2017-10-19
 remove nested comments that generate a compiler error using icc

rbertens:
nested comments are caught by the intel c++ compiler and treated as an error, it would be great if this (otherwise harmless) pr could be merged

example of compiler error:

/nics/c/home/rbertens/alibuild/sw/SOURCES/AliRoot/0/0/AD/ADrec/AliADDecision.cxx(226): error #9: nested comment is not allowed

compiler used

[rbertens@acf-login2 AliRoot]$ icc --version
icc (ICC) 17.0.2 20170213
Copyright (C) 1985-2017 Intel Corporation.  All rights reserved.


ecalvovi:
+1

PR #459, opened 2017-10-20
Fix for some compilers; problem with outcommented code: /* deleted

jluehder:
Some compilers seem to have Problems with /* inside of a comment


shahor02:
+1

PR #462, opened 2017-10-23
Update IB geometry to latest data

mario6829:
The IB geometry was updated to the latest data as requested at ITSU plenary session. In particular the metal stack was added to the Si sensor, the FPC aluminum layers are more realistic, and the glue between the chip and the sensor was added.


iouribelikov:
+1

PR #465, opened 2017-10-27
Add support for AliEn OCDB time machine

dberzano:
When using, for instance:

    AliCDBManager *cdb = AliCDBManager::Instance();
    cdb->SetDefaultStorage("raw://");
    cdb->SetRun(277015);
    cdb->SetMaxDate(TTimeStamp(2017, 7, 1, 0, 0, 0));  // new feature

only OCDB files created no later than the given UTC date will be returned.

This feature ensures perfect reproducibility: by setting a certain date it will
be possible to:

* protect jobs from OCDB updates while they are running
* re-run jobs in the future using the exact same OCDB

### DIFF
--- a/STEER/CDB/AliCDBGrid.cxx
+++ b/STEER/CDB/AliCDBGrid.cxx
@@ -1149,6 +1149,10 @@ void AliCDBGrid::QueryValidFiles()
 
   TString pattern = "Run*";
   TString optionQuery = "-y";
+  if(fMaxDate) {
+    optionQuery = "";  // AliEn find's `-y` returns no guid in fields :-(
+    AliInfoF("Returning only latest versions as they were at %s UTC", TTimeStamp(fMaxDate).AsString("s"));
+  }
   if(fVersion >= 0) {
     pattern += Form("_v%d_s0", fVersion);
     optionQuery = "";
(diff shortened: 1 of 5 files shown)


dberzano:
+test
fyi: @fprino @chiarazampolli @mconcas

dberzano:
(If you spot something very wrong in this PR could you please let me know?)

fprino:
Hi Dario,

thanks for implementing this feature.

I just have a question: from the diff of AliCDBGrid.cxx it seems you have also added a check on the version number of the file for files with GuidToCreationTimestamp(o_guid->String()) < fMaxDate).
I thought (but I well may be wrong) that this check should have been already applied in somewhere else in the code in order to return the latest version of the file, which is done also when fMaxDate is not used.
But I may be wrong...

Cheers, Francesco

dberzano:
Hi @fprino, so: the original code (without the time machine feature) used a special `alien_find` syntax supporting a `-y` flag. The AliEn query returns only the latest version of the OCDB files and this is implemented on the server side.

Since what I do does not have any server-side optimized version, I had to:

* make AliEn return all files,
* apply the "last version filter" in the C++ code.

So when you say:

> this check should have been already applied in somewhere else in the code in order to return the latest version of the file

you are right, but it is applied on the server side only.

Note that in case you don't use the "time machine" feature, the original AliEn find command is used with the original server-side version check optimization.

Does this clarify?

fprino:
Hi Dario,

thanks for the explanations. All clear now!
I did not know of the -y option of alien_find, sorry for the noise I created.

Cheers, Francesco

dberzano:
+1

No noise at all. This option is quite ad-hoc and it took me a while to figure out what it did, so it's good to discuss it and have it documented somewhere!

Can you please open a JIRA (as usual) in case you want the feature to be backported to v5-08-13ze-cookdedx?

PR #467, opened 2017-10-28
Adapt the trigger raw digit maker to work with raw stream in DATE format

aiola:
These changes are needed to use this class also to decode the STU data in the DQM framework.
Nothing changes when this class is used for reconstruction using the raw ROOT format.


aiola:
See discussion in this [JIRA ticket](https://alice.its.cern.ch/jira/browse/EMCAL-40)

mfasDa:
+test

Would be nice if you could have as commit message in case it refers to a JIRA ticket:
"[EMCAL-40] Factor out ..."
With this it makes it easier to keep track of changes.

mfasDa:
Hi @aiola ,

Thanks for the fix. Looks like you forgot the first commit message as well. Could you fix this as well?
Once this is done I will approve it.

Cheers

Markus

mfasDa:
+1

Thanks a lot! Based on previous discussions I approve the PR.

PR #469, opened 2017-10-31
Add Cint dependency to libAliHLTUtil.so

syamasak:
Hi,
This commit add Cint dep to AliHLTUtil because AliRoot build fails on macOS 10.13 with the following error. libCint has these missing symbols but it was not directly linked to AliHLTUtil.

```
[64%] Linking CXX shared library libAliHLTUtil.so
Undefined symbols for architecture x86_64:
  "_G__Doubleref", referenced from:
      G__G__AliHLTUtil_517_0_3(G__value*, char const*, G__param*, int) in G__AliHLTUtil.cxx.o
  "_G__Floatref", referenced from:
      G__G__AliHLTUtil_930_0_33(G__value*, char const*, G__param*, int) in G__AliHLTUtil.cxx.o
  "_G__UShortref", referenced from:
      G__G__AliHLTUtil_517_0_3(G__value*, char const*, G__param*, int) in G__AliHLTUtil.cxx.o
  "_G__add_setup_func", referenced from:
      __GLOBAL__sub_I_G__AliHLTUtil.cxx in G__AliHLTUtil.cxx.o
  "_G__call_setup_funcs", referenced from:
      __GLOBAL__sub_I_G__AliHLTUtil.cxx in G__AliHLTUtil.cxx.o
  "_G__check_setup_version", referenced from:
      _G__cpp_setupG__AliHLTUtil in G__AliHLTUtil.cxx.o
  "_G__defined_typename", referenced from:
      G__setup_memvarAliHLTAgentUtil() in G__AliHLTUtil.cxx.o
      G__setup_memfuncAliHLTAgentUtil() in G__AliHLTUtil.cxx.o
      G__setup_memvarAliHLTBlockFilterComponent() in G__AliHLTUtil.cxx.o
      G__setup_memfuncAliHLTBlockFilterComponent() in G__AliHLTUtil.cxx.o
      G__setup_memvarAliHLTCaloClusterReader() in G__AliHLTUtil.cxx.o
      G__setup_memfuncAliHLTCaloClusterReader() in G__AliHLTUtil.cxx.o
      G__setup_memvarAliHLTCompStatCollector() in G__AliHLTUtil.cxx.o
      ...
  "_G__double", referenced from:
      G__G__AliHLTUtil_930_0_27(G__value*, char const*, G__param*, int) in G__AliHLTUtil.cxx.o
      G__G__AliHLTUtil_930_0_28(G__value*, char const*, G__param*, int) in G__AliHLTUtil.cxx.o
[...]

PR #471, opened 2017-10-31
Fix ROOT6 build linker error introduced in #469

mfasDa:
In #469 an explicit dependency on libCint was added to
libHLTbase without checking the ROOT version. As libCint
does not exist for ROOT6 this produces a linker error.
This version distinguishes between ROOT5 and ROOT6 builds
using libCling instead of libCint for ROOT6 builds.

### DIFF
--- a/HLT/BASE/CMakeLists.txt
+++ b/HLT/BASE/CMakeLists.txt
@@ -147,8 +147,11 @@ set(HDRS ${HDRS}
 # It will create G_ARG1.cxx and G_ARG1.h / ARG1 = function first argument
 get_directory_property(incdirs INCLUDE_DIRECTORIES)
 generate_dictionary("${MODULE}" "${MODULE}LinkDef.h" "${HDRS}" "${incdirs}")
-
-set(ROOT_DEPENDENCIES Cint Core Geom Graf Hist MathCore Net RIO Tree XMLParser)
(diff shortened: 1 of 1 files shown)


mfasDa:
@mkrzewic @davidrohr Could you please confirm whether the dependency on the interpreter library is really necessary? Thanks!

Cheers

Markus

syamasak:
@mfasDa  Thanks!
BTW, ROOT5 and before uses Cint and future version of ROOT (at least 7) would use Cling. So, the following seems better. 
```
set(ROOT_DEPENDENCIES Core Geom Graf Hist MathCore Net RIO Tree XMLParser)

if(${ROOT_VERSION_MAJOR} LESS 6)
	list(APPEND ROOT_DEPENDENCIES Cint)
else()
	list(APPEND ROOT_DEPENDENCIES Cling)
endif(${ROOT_VERSION_MAJOR} LESS 6)
```
I don't know ROOT6 really needs Cling dependency..

mfasDa:
As the issue was reverted in #472 I close the pull request. @syamasak Indeed your version is more flexible. Maybe #469 can be implemented in this way.

mkrzewic:
Since only the cmake files are changed here only the offline team was notified. Why it this change needed?

PR #477, opened 2017-11-07
Fix missing member initialization

dberzano:
Feature introduced in #465 is missing a member initialization causing an
undetermined and very dangerous behaviour in most cases.


dberzano:
+test
@chiarazampolli @brinick @ktf @costing: yes, the problem on the SHUTTLE test node was very likely due to this bit.

For the record, log messages showed:

```
I-AliCDBGrid::QueryValidFiles: Returning only latest versions as they were at 1915-07-24 21:24:41 UTC
```

when `fMaxDate` was not set explicitly. The variable is now explicitly set to 0.

dberzano:
+1

PR #479, opened 2017-11-08
Fixed q. efficiency bug resulting in FIT.Hits overestimation, few obs…

mslupeck:
…olete lines removed, style fixed

Diff is big mostly due to style reformatting. The main real changes are:
- lines 961-966 - commenting out 3 function calls, which are not handled by GEANT3 - and only generate warnings during simulation - I have tested that this change does not influence the contents of FIT.Hits.root
-  1129 - actual bug fix


AllaMaevskaya:
I think mslupeck should commit not in master but fit-dev branch, then i will approve it and ask to merge with master myself.

PR #489, opened 2017-11-19
New X* state added

mpuccio:
Following a discussion with @amorsch it has been clarified that the Pythia decayer cannot handle the decay of di-baryon resonances like the d*. For this reason I am bringing back the modifications to the AliMC framework.


dberzano:
+1 (formal approval from @amorsch has been verified)

PR #498, opened 2017-11-29
Add new event selection features: by cluster and by triggered detector

fprino:
This commit introduces two new ways to pre-select events to be reconstructed in AliRawReader:
1) events can be selected based on the trigger cluster, which is extracted by parsing the string of the fired trigger cluster. E.g.
rec("raw.root","?Cluster=CENT") -> reconstruct events of the CENT cluster
rec("raw.root","?Cluster=CENT || CALO") -> reconstruct events of the CENT and CALO clusters
rec("raw.root","?ExcludeCluster=CENT") -> rejects events of the CENT cluster

2) events can be selected based on the presence/absence of specific detectors in the list of triggered detectors. E.g.
rec("raw.root","?DetectorsContain=ITSSDD") -> reconstruct events in which the SDD were triggered
rec("raw.root","?DetectorsExcluded=ITSSDD") -> reject events in which the SDD were triggered

These features are needed for the reconstruction of the pp reference wun of 2017 (LHC17p and LHC17q) to treat separately the events collected with SDD (CENT and CALO clusters) from those without SDD (= those that are in the FAST and CALOFAST clusters but not in the CENT and CALO).

### DIFF
--- a/RAW/RAWDatarec/AliRawReader.cxx
+++ b/RAW/RAWDatarec/AliRawReader.cxx
@@ -75,6 +75,10 @@ AliRawReader::AliRawReader() :
   fSelectTriggerExpr(),
   fContainTriggerExpr(),
   fExcludeTriggerExpr(),
+  fSelectClusterExpr(),
+  fExcludeClusterExpr(),
+  fSelectDetectorExpr(0),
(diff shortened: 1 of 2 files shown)


dberzano:
+1

PR #506, opened 2017-12-03
Internal GetTrack calls must go via top event

shahor02:
Within the AliMCEvent all internal calls of GetTrack with global track ID must
go via top MCEvent, since the global indices do not correspond to that in the
subsidiary event stack. In MCtoMC embedding mode this topEvent is the one containing
all subsidiaries, in standard MC topEvent points to itself.
W/o this feature AssignGeneratorIndex was crashing in MC embedding mode if 
injected generator was cocktail.


shahor02:
+1

PR #507, opened 2017-12-04
EvtGen: small bug fix

aiola:
Bug fix: nt was uninitialized and no check was done on the result of PushTrack leading to seg fault in KeepTrack when particle was not pushed (e.g. PDG not known)


dberzano:
+test

aiola:
Hello,
I do not know who is responsible for this class, however this is a rather trivial fix that is needed to avoid a crash whenever a particle not known to TDatabasePDG ends up in the decay products.
Best,
Salvatore

dberzano:
+1

PR #510, opened 2017-12-06
ATO-83, ATO-433: More verbose tree queries in case of failures

miranov25:
TStatToolkit.cxx, AliTreeTrending.cxx
  * ATO-83:  Protection against invalid tree queries
  * ATO-433: Protection against of failure in case not cut specified


miranov25:
+1

PR #512, opened 2017-12-06
New single injection per event mode

mpuccio:
Hi @amorsch ,
I added the modification to AliGenCocktail that we discussed yesterday.
The default behaviour does not change. The modifications are minimal but the fix of formatting makes them not very visible. Please check the diff without white spaces changes [here](https://github.com/alisw/AliRoot/compare/master...mpuccio:master?w=1) (or adding ?w=1 to the url of this page).
Cheers,
Max

### DIFF
--- a/EVGEN/AliGenCocktail.cxx
+++ b/EVGEN/AliGenCocktail.cxx
@@ -18,9 +18,9 @@
 // Container class for AliGenerator through recursion.
 // Container is itself an AliGenerator.
 // What is stored are not the pointers to the generators directly but to objects of type
-// AliGenCocktail entry.   
-// The class provides also iterator functionality.  
-// Author: andreas.morsch@cern.ch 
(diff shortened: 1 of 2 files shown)


rbertens:
+1 thanks @mpuccio  for providing the diff without white space changes

amorsch:
+1

miweberSMI:
Hi @mpuccio, all,

could you just quickly explain, what the new functionality does? I assume you produce one particle per event, but from a larger "pool" of particles?

Thanks and cheers,
Michael

mpuccio:
Ciao @miweberSMI ,

with this feature the first generator in the cocktail is considered as underlying event while all the others are the signal generators. For each event, enabling the SingleInjectionPerEvent flag, it generates the underlying event and one and only one of the other generators picked randomly.

Cheers,
Max

PR #513, opened 2017-12-06
Updates of HLT TPC CA Tracker, CA Tracking standalone QA, O2 Tracking

davidrohr:


dberzano:
Every CMake there points one level up: this is an ugly practice, in fact the CMakeLists.txt should be moved one level up and pointers refactored accordingly. I know this hasn’t been done in the past but I think it needs to be corrected once we’re here...

dberzano:
+1 Follow-up from a discussion with @davidrohr: we'll allow this for now and refactor later.

PR #514, opened 2017-12-08
kMuonCalo alias added

ekryshen:


dberzano:
+test - may you please provide us with a reference proving that this is known and has been discussed somewhere? I see that it is not altering any data structure but just adding an alias, but still it's better if some explanation is provided :-)

ekryshen:
For reference, I copy-paste my e-mail exchange with Diego Stocco:

Hi Diego,

Yes, I can try to add them. As usual we have a deficit of bits but I think I can reserve BIT(29) for this purpose. Would kMuonCalo alias be ok for this purpose?

Note that I would have to modify AliVEvent.h to introduce new bit so one would have to wait for the new aliroot tag.

Cheers,
Evgeny

dberzano:
+1 and thanks for reporting the email exchange!

PR #517, opened 2017-12-09
ATO-46, PWGPP-205- Bug fix

miranov25:
ATO-46          - Add index name for the Monalisa production table
PWGPP-205  - Check tree query buffer pointer


miranov25:
+1

PR #520, opened 2017-12-11
 ATO-434, ATO-83: FIX - warnings as reported by Marco 

miranov25:
not sure which  compiler used my Marco 

Some of the warnings should be default in the future
FIX: STAT/TStatToolkit.cxx
* [-Wlogical-not-parentheses]
* [-Wparentheses]


miranov25:
+1

PR #529, opened 2017-12-15
Using decayer in Init, when set

miweberSMI:
- if not set, use external decayer from TVirtualMC::GetMC() (as before)
- if no decayer available, return Fatal
- needed for adding Exodus decayer for single AliGenParam sources in cocktail

### DIFF
--- a/EVGEN/AliGenParam.cxx
+++ b/EVGEN/AliGenParam.cxx
@@ -438,7 +438,12 @@ void AliGenParam::Init()
 {
   // Initialisation
 
-  if (TVirtualMC::GetMC()) fDecayer = TVirtualMC::GetMC()->GetDecayer();
+  if (!fDecayer && TVirtualMC::GetMC()) fDecayer = TVirtualMC::GetMC()->GetDecayer();
+  if (!fDecayer){
+	Fatal("AliGenParam",
+              "No decayer attached");
+  }
+
   //Begin_Html
   /*
     <img src="picts/AliGenParam.gif">


miweberSMI:
Hi @amorsch, @qgp,

I created now the pull request that I mentioned in an earlier mail. Please have a look and let me know, if you agree with the proposed changes. It would be needed for a MC production from PWG-DQ that is soon to be proposed.

Thanks and cheers,
Michael

amorsch:
Ok with me.
However, in general you have to be careful having several instances of AliDeacyerPythia6
Behind the scenes it accesses the same FORTRAN COMMON blocks. Hence the instances 
will be not independent. For w/wo Exodus this should be OK.

amorsch:
+1

rbertens:
+1  Hi @amorsch , I don't think you're in the list or 'approvers' for this PR (just myself and @qgp , should this be changed?) In any case, hereby the +1 for merging. Have  a good weekend,
Redmer

miweberSMI:
Hi @amorsch,

thanks a lot. Could you tell me, what could checked in order to see that there is no problem?
What we want to do is using standard Pythia decayer for underlying HIJING events, but Exodus for injected AliGenParam.

Thanks 
Michael

amorsch:
Using the pythia native decayer and exodus at the same time is ok since Exodus does not make use of pythia common block. In all other cases one has to make sure that settings used for AliGenParam
do not interfer with those of the external decayer used during transport. Andreas.

miweberSMI:
OK, thanks for explanation.
Any idea, how one could prevent such misusage?

Thanks 
Michael

PR #538, opened 2017-12-24
 avoid floating point error when pz is larger than e

rbertens:
In the old condition, the Log((e+pz)/(e-pz)) is not evaluated when e == pz within float precision, but it is evaluated when pz > E (which is unphysical ,but can happen within float precision). By removing the outer TMath::Abs from the condition, pz > E is also caught, and will return a -999 as rapidity. Anyone who approves, please check if I didn't make a dumb mistake in logic ... thanks, Redmer


shahor02:
+1

PR #544, opened 2018-01-15
Speed up merge with root6

dstocco:
The merging of an AliMergeableCollection is extremely slow on ROOT6. This happens because the interpreter was used during merging. The PR uses TMethodCall instead, solving the speed issue.

### DIFF
--- a/STEER/STEERBase/AliMergeableCollection.cxx
+++ b/STEER/STEERBase/AliMergeableCollection.cxx
@@ -51,6 +51,7 @@ ClassImp(AliMergeableCollection)
 #include <cassert>
 #include <vector>
 #include "TBrowser.h"
+#include "TMethodCall.h"
 
 //_____________________________________________________________________________
 AliMergeableCollection::AliMergeableCollection(const char* name, const char* title) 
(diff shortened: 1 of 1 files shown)


dberzano:
+test does this work with ROOT5?

dstocco:
Ciao Dario,
yep, I tested it with ROOT5 as well and it works fine.

By the way: FYI, I recently tried again to install ROOT5 in High-Sierra. The ctests fail, but once removed I can compile + work with aliroot (included the graphical part). That's very good for debugging ;)

Ciao,
Diego

dberzano:
Seems there was a temporary error in the guntest (it eventually passed). Can you check the full log and see if this is related to your change?

dstocco:
Ciao Dario,
I see no error on the full log (just a warning which is not related to my modification). All of the library tests passed as well.

Cheers,
Diego

dstocco:
Ciao @dberzano 
it looks like all tests passed w/o any specific issue related with my commit. Do you see some further problems? If not, can we go with a +1?

Thanks!
Ciao,
Diego

dberzano:
+1 Hmm the log was replaced by... no log, so we have no more trace of the original error. Anyway let's proceed.

PR #547, opened 2018-01-16
fix weird ununderstandable problem with ROOT6 cling clashing in interpreter mode when cout is used in aliroot binary

davidrohr:
This circumvents a crash in ROOT 6 cling on some systems.
We have been trying to debug this with @pzhristov but could not find the real cause.


dberzano:
+1 weird... let's accept the workaround for now

PR #551, opened 2018-01-25
Silence a -Werror on GCCv7

dberzano:


dberzano:
+1 This PR makes AD compile correctly with GCCv7's default options. GCCv7 is the default compiler on Ubuntu 17.10 and 18.04 for instance.

dberzano:
+1 (added a check on the availability of `-Wno-error=misleading-indentation`)

PR #552, opened 2018-01-25
Introduce fast Dipole field parametrizations

syamasak:
@shahor02 Could you check the code? The dipole part should return exactly same results against AliMagF's.

### DIFF
--- a/STEER/STEERBase/AliMagFast.h
+++ b/STEER/STEERBase/AliMagFast.h
@@ -5,12 +5,15 @@
 
 //
 // Fast polynomial parametrization of Alice magnetic field, to be used for reconstruction.
-// Solenoid part fitted by Shuto Yamasaki from AliMagWrapCheb in the |Z|<260Interface and R<500 cm 
+// Solenoid part fitted by Shuto Yamasaki from AliMagWrapCheb in the |Z|<260Interface and R<500 cm
 // Dipole part: to do
 //
 // Author: ruben.shahoyan@cern.ch
 //
+#include <string>
+#include <math.h>
 #include <TObject.h>
+#include <TMutex.h>
 
 class AliMagFast : public TObject
 {
@@ -18,17 +21,37 @@ class AliMagFast : public TObject
  public:
   enum {kNSolRRanges=5, kNSolZRanges=22, kNQuadrants=4};
   enum {kX,kY,kZ};
-  
+
   struct SolParam { float mParBxyz[3][20];};
   typedef SolParam SolParam_t;
 
-  AliMagFast(const char* inpFName=0);
-  AliMagFast(Float_t factor, Int_t nomField = 5, const char* inpFmt="$(ALICE_ROOT)/data/maps/sol%dk.txt");
+  // For Dipole
+  struct SegmentEnd { UShort_t index; float endPos; };
+  typedef SegmentEnd SegmentEnd_t;
+  struct SegmentSearch {
+    int nDivision;
+    float factor;
+    float offset;
+    SegmentEnd_t *slices;
+    SegmentSearch *segments;
+  };
+  typedef SegmentSearch SegmentSearch_t;
+  struct ChebFormula {
+    float (*bz)(const float *xyz);
+    void (*bxyz)(const float *xyz, float *b);
+  };
+  typedef ChebFormula ChebFormula_t;
+
+  AliMagFast(Float_t factorSol=1.f, Float_t factorDip=1.f, Int_t nomField = 5,
+    const char* inpFmt="$(ALICE_ROOT)/data/maps/sol%dk.txt",
+    const char* inpFmtDip="libAliMagFastDip%dk",
+    const char* symFmtDip="dip%dk");
+  AliMagFast(const char* inpFName, const char* inpFNameDip, const char* symFmtDip);
   AliMagFast(const AliMagFast& src);
   AliMagFast& operator=(const AliMagFast& src);
-  
-  Bool_t LoadData(const char* inpFName);
-  virtual ~AliMagFast() {}
+
+  Bool_t LoadData(const char* inpFName, const char* inpFNameDip, const char* symDip);
+  virtual ~AliMagFast();
 
   Bool_t Field(const double xyz[3], double bxyz[3]) const;
   Bool_t GetBz(const double xyz[3], double& bz)     const;
@@ -37,10 +60,12 @@ class AliMagFast : public TObject
 
   void    SetFactorSol(float v=1.f)                       {fFactorSol = v;}
   Float_t GetFactorSol()                            const {return fFactorSol;}
-  
+  void    SetFactorDip(float v=1.f)                       {fFactorDip = v;}
+  Float_t GetFactorDip()                            const {return fFactorDip;}
+
  protected:
 
-  Bool_t GetSegment(const float xyz[3], int& zSeg,int &rSeg, int &quadrant) const;  
+  Bool_t GetSegmentSol(const float xyz[3], int& zSeg,int &rSeg, int &quadrant) const;
   static const float fgkSolR2Max[kNSolRRanges];       // Rmax2 of each range
   static const float fgkSolZMax;                      // max |Z| for solenoid parametrization
 
@@ -50,12 +75,18 @@ class AliMagFast : public TObject
     // get point quadrant
     return y>0 ? (x>0 ? 0:1) : (x>0 ? 3:2);
   }
-  
+
   float CalcPol(const float* cf, float x,float y, float z) const;
+  Bool_t QuickSearch(const SegmentSearch_t ss, const float z, UShort_t &id) const;
+  Bool_t GetSegmentDip(const float xyz[3], UShort_t &formulaId) const;
 
   Float_t fFactorSol; // scaling factor
   SolParam_t fSolPar[kNSolRRanges][kNSolZRanges][kNQuadrants];
-  
+  Float_t fFactorDip; // scaling factor
+  SegmentSearch_t fDipSegments;
+  ChebFormula_t *fDipPar;
+  std::string fLibNameDip;
+
   ClassDef(AliMagFast,1)
 };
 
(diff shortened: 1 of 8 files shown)


shahor02:
Hi @syamasak 
Thanks! Will check asap but this may take some time, rigth now too overloaded, sorry

syamasak:
@shahor02 I'm not in a hurry, take it easy.

syamasak:
@syamasak There are 3 problems to be fixed later
- Thread safety
- Forgot to multiply fFactorDip
- COMPILE_FLAGS seems not working as expected (Fused multiply add instructions should be used for the best performance)

dberzano:
+test

shahor02:
@dberzano : please don't merge this PR it for the moment, I am in the process of its validation

syamasak:
@dberzano @shahor02 Sorry, I found a severe bug in the impl. Please do not merge. I'm investigating it now.

dberzano:
@shahor02 of course, I have just approved the tests (not the merge) to speed up the turnaround in case you haven't seen it 👍

dberzano:
While we wait for the fix, the PR checker points out that the files you modify in this PR exceed 20 MB(!!!) in size. Moreover, I see that more libraries are added under STEERbase. This is not the convention we use in AliRoot: a _separate_ folder with a _separate_ `CMakeLists.txt` has to be added to hold the new libraries. Before proper restructuring, this PR must not be approved.

Thanks!

syamasak:
@davidrohr I'm sorry for commiting such large files. They come from (generated) parametrizations of B-field (about 4 times faster compared to AliMagF). They are compiled into separate libraries and dynamically loaded at runtime whenever needed. I agree putting generated source is not ideal way, but generating these parametrizations take some time. I'm willing to implement any alternative idea if any, welcome!

> This is not the convention we use in AliRoot:
I think that moving these parametrizations into STEER/STEERBase/AliMagFastDip{2,5}k/ should be ok, right?

Thanks!

davidrohr:
Hi @syamasak , probably you wanted to ping @dberzano and not me.

Cheers
David

shahor02:
Hi Shuto, we can seperate it later to special directory, let me 1st check the code.
Cheers,
 Ruben

syamasak:
Hi @davidrohr, you are right! I am sorry to bother you.

Best regards,
Shuto

syamasak:
Hi Ruben, current dipole parametrization seems return random values in some area. I didn't noticed it when I created the PR since it worked with small testcases (e.g. (0,0,-1000)). Maybe something went wrong in GetSegmentDip().

Cheers
Shuto

dberzano:
@syamasak yes the proposed structure works for us :-)

How large are those data files in total? How frequently are they expected to change, in the future? The reason why we don't accept them is that a Git repository grows in size over time, and we'd like to keep this growing under control.

So, _if_ these files, once generated, are never going to change again, and _if_ their size is reasonable, what we can do is having two pull requests:

* We keep this one for the _code_ only (which is slim) in order to allow the checkers to do their job. Checkers refuse to run if the PR size is too large as you have seen.
* We make another separate PR for data only. Even if it exceeds the 20 MB size, if the conditions above are satisfied we can force-merge it without the check.

OK?

syamasak:
@dberzano This two dipole parametrizations are generated from $(ALICE_ROOT)/data/maps/mfchebKGI_sym.root and never chenge unless the electromagnetic field of ALICE is measured again.
As current commit contains many spaces and redundant names in it, I think that I can make it slimmer.

dberzano:
Hi, what is the size of such ROOT file? It is pointless to make source code slimmer, the limit is 20 MB and a few bytes won't make any difference at all.

syamasak:
Hi, sorry for bad explanation.
Above ROOT file seems commited long time ago and does not increase repo size anymore.
The data files of this commit are actually generated C++ source code of 466273 lines in total. Please take a look at this:
<https://raw.githubusercontent.com/syamasak/AliRoot/6a48d7ffb68d0a76d583a6bf3274f26a5819a47c/STEER/STEERBase/AliMagFastDip2k.cxx>

Stripping spaces and float point numbers with too much digits in it can reduce many bytes.

Thanks!

dberzano:
I see. Given it's a text file, it's OK (Git compresses it efficiently). In order to let the PR pass, I still suggest you to create a _different_ PR with this file only (we'll approve it manually), and then you keep the functionality in this PR.

shahor02:
Hi Shuto,
The build fails for 2 reasons: typo in the cmake and using unordered_map which require c++11.
For the 2nd problem: we had very bad experience (memory corruption) in the past with unloading root libraries and I don't see in which conditions we will need to change the field from 5 to 2 kG in the same session. So, at the moment I've simply suppressed the static counters/containers for the loaded field libs. Instead, I pushed the fixes in my fork of your branch: https://github.com/shahor02/AliRoot/commits/feature-fast-dipole, see commits https://github.com/shahor02/AliRoot/commit/3864017384bf7844b5ef87c2629ae4500cc00a89 and https://github.com/shahor02/AliRoot/commit/449c7262504552ce1fcac58b9ddba527393c551a.
For the safety, you can consider having a static flag set when the field lib is loaded and produce a fatal on the attempt to load another library.

Cheers,
 Ruben

shahor02:
+test

On Mar 16, 2018 12:49, ALICE Builder <notifications@github.com> wrote:

247feaf<https://github.com/alisw/AliRoot/commit/247feaf0f137a39d667cd43280c3b5dcd54f7de2>: approval required: 1 of @jgrosseo<https://github.com/jgrosseo> (Jan Fiete Grosse-Oetringhaus), @chiarazampolli<https://github.com/chiarazampolli> (Chiara Zampolli), @qgp<https://github.com/qgp> (Jochen Klein), @dberzano<https://github.com/dberzano> (Dario Berzano), @sawenzel<https://github.com/sawenzel> (Sandro Christian Wenzel), @ktf<https://github.com/ktf> (Giulio Eulisse), @shahor02<https://github.com/shahor02> (Ruben Shahoyan)

Comment with +1 to approve and allow automatic merging,or with +test to run tests only. Please comment on the pull request: click here<https://github.com/alisw/AliRoot/pull/552> and comment at the bottom of the page.

STEER/STEERBase/AliMagFastDip5k/AliMagFastDip5k.cxx
```diff
@@ -0,0 +1,11 @@
+#include "../AliMagFast.h"
```
shahor02: AliMagFast is common header and since in the aliroot installation we store all headers in the common $ALICE_ROOT/include/, this relative path may create problems.

syamasak:
@shahor02 The header is intentionally blank and I'll push the content in another PR.

STEER/STEERBase/AliMagFastDip5k/AliMagFastDip5k.cxx
syamasak: Perhaps #include <AliMagFast.h> seems better?

shahor02: I would say #include "AliMagFast.h"

shahor02:
@syamasak But AliMagFast is used in the DPG productions (with fall-back to slow field for dipole part). If the real headers come from different PR, in between the AliMagF will not be operational?

syamasak:
@shahor02 Modified so that AliMagFast::QuickSearch always return kFALSE.

syamasak:
@shahor02 since QuickSearch return kFALSE, GetSegmentDip return kFALSE and AliMagFast::Field return kFALSE. In other words, AliMagFast behaves as if dipole part was not exist.

STEER/STEERBase/AliMagFastDip5k/AliMagFastDip5k.cxx
```diff
@@ -0,0 +1,11 @@
+#include "AliMagFast.h"
+#include "AliMagFastDip5k.h"
+typedef unsigned short ushort;
+const float Infinity = INFINITY;
+AliMagFast::SegmentEnd_t dip5k_z_slices[] = {
+};
+AliMagFast::SegmentSearch_t dip5k_z_segments[] = {
+};
+AliMagFast::SegmentSearch_t dip5k_z = {-1, 0.0, -1760.0, dip5k_z_slices, dip5k_z_segments};
```
shahor02: @syamasak building on ubuntu I get an error here: 
````
/home/shahoian/alice/sw/SOURCES/AliRoot/0/0/STEER/STEERBase/AliMagFastDip5k/AliMagFastDip5k.cxx:9:90: error: narrowing conversion of ‘-1’ from ‘int’ to ‘UShort_t {aka short unsigned int}’ inside { } [-Wnarrowing]
 AliMagFast::SegmentSearch_t dip5k_z = {-1, 0.0, -1760.0, dip5k_z_slices, dip5k_z_segments};
                                                                                          ^
/home/shahoian/alice/sw/SOURCES/AliRoot/0/0/STEER/STEERBase/AliMagFastDip2k/AliMagFastDip2k.cxx:9:90: error: narrowing conversion of ‘-1’ from ‘int’ to ‘UShort_t {aka short unsigned int}’ inside { } [-Wnarrowing]
 AliMagFast::SegmentSearch_t dip2k_z = {-1, 0.0, -1760.0, dip2k_z_slices, dip2k_z_segments};
                                                                                          ^
````
Indeed, the SegmentSearch_t is defined as 
````
  struct SegmentSearch {
    UShort_t nDivision;
    float factor;
    float offset;
    SegmentEnd_t *slices;
    SegmentSearch *segments;
  };
````
I guess the initialization should be done as  ````dip5k_z = {0, ...````

File: STEER/STEERBase/AliMagFast.h [hunk 3]
```diff
@@ -71,9 +102,31 @@ inline float AliMagFast::CalcPol(const float* cf, float x,float y, float z) cons
...
     z*(cf[3] + z*(cf[9] + x*cf[15] + y*cf[18] + z*cf[19]) + x*(cf[6]) );
-    
+
   return val;
 }
 
+inline Bool_t AliMagFast::QuickSearch(const SegmentSearch_t ss, const float z, UShort_t &id) const
+{
+  const int index = floor((z - ss.offset) * ss.factor);
+  if (index > ss.nDivision || index<0) return kFALSE;
```
shahor02: Related to previous: If the nDivision is indeed the number of divisions, should not you use here
````
if (index >= ss.nDivision || index<0) return kFALSE;
````
Also, to avoid ushort - int comparison, I would use int for nDivision

syamasak: Sorry, wait a moment

shahor02: So, do you keep here ````index > ss.nDivision````?

syamasak: index can be any integer including negative number.
So `index < 0` is necessary. `(so index > -1 && index < 0)` is alway false.

syamasak: OK, local build succeeded!

shahor02: @syamasak I mean is index==ss.nDivision valid? From the following code I would say no.

syamasak: What a basic mistake... 
// Why humans should take care such kind of mistake? (because the compiler is not smart enough.)

Any other bug or suggestion? I'll make the PR for the data.

shahor02: @syamasak so, if it should be ````index >= ss.nDivision````, would you fix it? 
The rest looks fine.

syamasak: @shahor02 Oh, the fix has not staged correctly. I push it now.

syamasak: @shahor02 this version seems ok. Thank you!

shahor02:
+1

syamasak:
@shahor02 I appreciate your kindness.

PR #564, opened 2018-02-01
T0 dev T0A+ and V0A+ support structure are included

AllaMaevskaya:


AllaMaevskaya:
Hello

how can I repeat this crash? This commits are about FIT detector , it is not inside Run2 geometry and should not be used with usual Config.C but upgrade one

Best regards

Alla

PR #575, opened 2018-02-12
Automatise setting of pass name for MC

wiechula:
* Read missing pass name or overwrite if it exists from OADB
* Check for multiple pass names for one production and check if this is
handled by the OADB
* Force pass name from AliProdInfo if it exists


chiarazampolli:
+1

PR #577, opened 2018-02-14
change angle from radian to degrees for shower shape calculation; in …

gconesab:
…case of negative fw0 just pass as weight the log ecluster/ecell


gconesab:
+1

PR #581, opened 2018-02-20
Fixes for ROOT6 

chiarazampolli:
The fitting in ROOT6 needs some initializations that are not necessary with ROOT5. This fix was done together with @shahor2.

### DIFF
--- a/TPC/TPCcalib/AliTPCcalibAlign.cxx
+++ b/TPC/TPCcalib/AliTPCcalibAlign.cxx
@@ -715,10 +715,13 @@ void  AliTPCcalibAlign::ExportTrackPoints(AliVEvent *event){
     AliTPCseed *seed0 = 0,*seed1=0;
     AliTPCseed tpcSeed0, tpcSeed1;
     //
+    Int_t nFriendTracks = Vfriend->GetNumberOfTracks();
+    if (index0 > nFriendTracks) continue; // most likely, we didn't save the friends for this event
     friendTrack = const_cast<AliVfriendTrack*>(Vfriend->GetTrack(index0));
     if (!friendTrack) continue;
     if (friendTrack->GetTPCseed(tpcSeed0)==0) seed0=&tpcSeed0;
     if (index1>0){
+      if (index1 > nFriendTracks) continue; // most likely, we didn't save the friends for this event
       friendTrack = const_cast<AliVfriendTrack*>(Vfriend->GetTrack(index1));
       if (!friendTrack) continue;
       if (friendTrack->GetTPCseed(tpcSeed1)==0) seed1=&tpcSeed1;
--- a/STAT/TStatToolkit.cxx
+++ b/STAT/TStatToolkit.cxx
@@ -431,6 +431,15 @@ TGraphErrors * TStatToolkit::MakeStat1D(TH2 * his, Int_t deltaBin, Double_t frac
     TH1 *projection = his->ProjectionY(name,TMath::Max(jx-deltaBin,1),TMath::Min(jx+deltaBin,nbinx));
     Double_t stat= 0;
     Double_t err =0;
+    if (projection->Integral()==0) {
+      vecX[icount] = xcenter;
+      vecY[icount] = stat;
+      vecYErr[icount] = err;
+      icount++;
+      delete projection;
+      continue;
+    }
+
     TStatToolkit::LTMHisto((TH1F*)projection,vecLTM,fraction);  
     //
     if (returnType==0) {
@@ -464,6 +473,10 @@ TGraphErrors * TStatToolkit::MakeStat1D(TH2 * his, Int_t deltaBin, Double_t frac
       const Int_t    maxBin = projection->GetMaximumBin();
       const Double_t max    = projection->GetXaxis()->GetBinCenter(maxBin);
       const Double_t range  = fraction*(projection->GetXaxis()->GetXmax()-projection->GetXaxis()->GetXmin());
+      f1.SetParameters(projection->GetMaximum(),
+		       projection->GetMean(),
+		       projection->GetRMS());
+      f1.SetRange(max-range, max+range);
       projection->Fit(&f1,"QN","QN", max-range, max+range);
       stat= f1.GetParameter(1);
       err=f1.GetParError(1);
--- a/STEER/ESD/AliKFParticle.h
+++ b/STEER/ESD/AliKFParticle.h
@@ -329,7 +329,8 @@ class AliKFParticle :public AliKFParticleBase
 
   //* Method to access ALICE field 
  
-  static Double_t GetFieldAlice();
+  static Double_t GetFieldAlice() { return fgBz;}
+
   
   //* Other methods required by the abstract AliKFParticleBase class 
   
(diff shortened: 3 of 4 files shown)


chiarazampolli:
+1

TPC/TPCcalib/AliTPCcalibAlign.cxx
```diff
@@ -715,10 +715,13 @@ void  AliTPCcalibAlign::ExportTrackPoints(AliVEvent *event){
     AliTPCseed *seed0 = 0,*seed1=0;
     AliTPCseed tpcSeed0, tpcSeed1;
     //
+    Int_t nFriendTracks = Vfriend->GetNumberOfTracks();
+    if (index0 > nFriendTracks) continue; // most likely, we didn't save the friends for this event
```
mkrzewic: ciao, wouldn't a better check be: if (ntracks != nFriendTracks) ? Or can we have more tracks than friend tracks while having a non-zero number of friend tracks?

TPC/TPCcalib/AliTPCcalibAlign.cxx
```diff
@@ -715,10 +715,13 @@ void  AliTPCcalibAlign::ExportTrackPoints(AliVEvent *event){
...
     friendTrack = const_cast<AliVfriendTrack*>(Vfriend->GetTrack(index0));
     if (!friendTrack) continue;
     if (friendTrack->GetTPCseed(tpcSeed0)==0) seed0=&tpcSeed0;
     if (index1>0){
+      if (index1 > nFriendTracks) continue; // most likely, we didn't save the friends for this event
```
mkrzewic: same comment as above

TPC/TPCcalib/AliTPCcalibAlign.cxx
chiarazampolli: Hello,

Probably you are right, I am quite sure that @shahor02 could reply with 100% certainty :) 

Cheers,

Chiara

shahor02: @mkrzewic : since LHC15o we are storing maximum ````AliReconstruction::fMaxFriendTracks (def=3000)```` tracks, see https://alice.its.cern.ch/jira/browse/ALIROOT-6411?focusedCommentId=161468&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-161468 for details. So, in high mult events we can have ````0<Nfriends<Ntracks````

mkrzewic: Hi Ruben, thanks. Then, if I understand it correctly the check should be `if (esdTrack->GetFriendNotStored())`, right?

mkrzewic: ...or better yet, just check for null pointer as the friends are attached to tracks based on criteria other than just the index.

chiarazampolli: The pointer is already checked but there was an error - now I don't recall the details, but I can reproduce it back.

haavardh:
+1

TPC/TPCcalib/AliTPCcalibAlign.cxx
shahor02: Actually, the correct way (compatible with pre-2015 and after data) to get the friends is AliESDtrack->GetFriendTrack(), which will return null if this particular friend was not stored.

mkrzewic: exactly, that is the problem. for code that also needs to run online on the HLT this would not work.

TPC/TPCcalib/AliTPCcalibAlign.cxx
mkrzewic: Hi @shahor02, I don't see how that would be dangerous? I thought the friend tree has to have the same number of entries to preserve the matching - so some of them will be empty and you get a null pointer for those cases, the same as when calling track->GetFriendTrack() (which just falls back to reading the same entry from the friend tree/branch).

chiarazampolli: Hello,

Meanwhile Haavard approved the PR, so we will need a new commit in case we want to change this.

Chiara

mkrzewic: I think we will want to change this, best to have a TPC expert look at this, e.g. @miranov25

chiarazampolli: Ciao Mikolaj.
For me it is fine to change it, for now this works, even if it is probably not the most elegant. 
Chiara

shahor02: Hi, today I've fixed many lines like that, but it seems missed this one. What I've changed is from
````
AliVTrack *track = event->GetVTrack(i);
...
friendTrack = const_cast<AliVfriendTrack*>(Vfriend->GetTrack(i));
````
to 
````
AliVTrack *track = event->GetVTrack(i1);
AliVfriendTrack *friendTrack = 0;
if (track->IsA()==AliESDtrack::Class()) {
   friendTrack = (AliVfriendTrack*)(((AliESDtrack*)track)->GetFriendTrack());
}
else {
   friendTrack = const_cast<AliVfriendTrack*>(vFriend->GetTrack(i));
}
````

Shall I change this file too?

haavardh:
Sorry -- I probably acted too quickly. As none of my code was affected, it was all OK from my side. I see I should have refrained from doing anything in this case.

Sorry for the confusion. I learned something for the next iteration....

                                                                                                                              - Håvard

TPC/TPCcalib/AliTPCcalibAlign.cxx
mkrzewic: Hi Ruben, sure, go ahead, but we were hoping to avoid these kinds of checks (again introducing different code paths based on input type). In principle we could think about introducing something like this in the V interface and flat ESD, but that might be too much work - including some CPU overhead due to the need to explicitly attach some pointers/indexes to the tracks (happens in the AliAnalysisManager for offline, always introducing this overhead). I think index based access to the friend tree directly is a much more elegant solution.
Now about the check: it is wrong algorithmically, it will possibly bias the result as it will only process the friends attached to the first 3000(fMax...) tracks only. Bias might be small, but still... :)

shahor02: @mkrzewic : I agree about the indixed access, but for historical reasons we have to support the esdTrack->GetFriendTrack method. One can access sparse friends also via indices:
````
int idx = esdTrack->GetFriendTrackID();
AliESDfriendTrack* frTrack = idx<0 ? 0 : esdFrEvent->GetTrack(idx);
````

chiarazampolli: @mkrzewic : "Now about the check: it is wrong algorithmically, it will possibly bias the result as it will only process the friends attached to the first 3000(fMax...) tracks only. Bias might be small, but still... :)" --> 
Actually also the current code:

friendTrack = const_cast<AliVfriendTrack*>(Vfriend->GetTrack(index0));

is wrong, as it is not guaranteed that the index of a track is not the same as the index of its friend.

shahor02: done as https://github.com/alisw/AliRoot/pull/593

mkrzewic: the joy of having friends...

chiarazampolli: wait until we include enemies in AliRoot... :)

dberzano: LOL

PR #584, opened 2018-02-20
vdt sincos with arg>1.6e9 mapped to libm sin,cos

shahor02:
vdt sin,cos, sincos require argument within MAX_INT*pi/4 range


chiarazampolli:
Hello @shahor02 ,
I would propose to open a JIRA ticket (porting request) for this change, would you agree? So to have it tracked for v5-09-24.

Cheers,

Chiara

PR #590, opened 2018-02-23
Commenting prints and putting a protection for undefined pc in NextParticle 

coppedis:
Despite the large number of lines tagged as changed, in this commit I simply:
1) commented a warning print that was useless (but annoying)
2) most importantly, put a protection in NextParticle() method that forbid not well defined particle to be propagated to the stack for tracking.

I also would like to explain that this external reader generator reads ntuples from RELDIS generator (dedicated to EM dissociation events) and that it is NOT used if not by myself for local productions since the RELDIS ntuples are not widely distributed.

I thank in advance whoever will allow this changes to be ported to the release.
Regards, ch.


rbertens:
+1

PR #592, opened 2018-02-27
Possibility to use multiple contain selections on trigger class

fprino:
This commit allows to select trigger classes using multiple selection of type "contain", which are treated in OR. 
For example one can run the rec.C with the option "?Contain=CCUP || CTRUE" to reconstruct the events of all the trigger classes CTRUE* and CCUP*.
With this feature, in the future, reconstructions of specific triggers should be easier to configure.
For example, the LHC15o reprocessing of ALIROOT-7611, which required 2 separate cycles, can be done in one cycle with the modified code.

### DIFF
--- a/RAW/RAWDatarec/AliRawReader.cxx
+++ b/RAW/RAWDatarec/AliRawReader.cxx
@@ -338,12 +338,9 @@ AliRawReader* AliRawReader::Create(const char *uri)
 	continue;
       }
       else if((option.BeginsWith("Contain=",TString::kIgnoreCase))){
-	if(contExpr.IsNull()){
-	  option.ReplaceAll("Contain=","");
-	  contExpr=option.Data();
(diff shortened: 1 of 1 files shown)


dberzano:
Hi @fprino, can you please fix the spaces in your code? Such as `for[space](...)[space]{`. Thanks!

fprino:
Hi @dberzano,

should I add the spaces in the for loops in the lines of code that I added in this commit?
Or do you want me to change it for all the for loops in the class (I had a quick look and there are many for loops without the [space] around the parentheses.

Cheers, Francesco

dberzano:
+1 @fprino you are right, I only had a look at the diff. I don't think it's worth fixing it in any case by now (we've never enforced any coding convention like this in AliRoot) so I am approving it as it is.

PR #594, opened 2018-02-28
Add the possibility of injecting single particle species via macro configuration

mpuccio:
This PR adds a simple way to configure AliGenParam to inject a particular particle species directly from an interpreted macro.
Due to a limitation of CINT it is not possible to pass an interpreted function to a library function, meaning that any new injection scheme would require adding a piece of code in AliRoot/AliPhysics instead of simply hacking the AliDPG repository.
This makes easier to replace AliGenBox with AliGenParam since now the functionality is the same with this new constructor.

@akalweit

### DIFF
--- a/EVGEN/AliGenParam.cxx
+++ b/EVGEN/AliGenParam.cxx
@@ -72,7 +72,8 @@ AliGenParam::AliGenParam()
   fForceConv(kFALSE),
   fKeepParent(kFALSE),
   fKeepIfOneChildSelected(kFALSE),
-  fPreserveFullDecayChain(kFALSE)
+  fPreserveFullDecayChain(kFALSE),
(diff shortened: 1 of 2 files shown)


rbertens:
+test

qgp:
Even with CINT it should be possible to compile (!) a function and pass it to the constructor, no?

Anyway, this appears to be a simple solution to do what you need; so, fine with me.

Cheers,
   Jochen

rbertens:
+1

rbertens:
+1

dberzano:
For some reason, out "C++98" test passes, but the native build on SLC6 (still - unfortunately - required for producing the DA RPMs). I'll fix it (I believe that `EVGEN` can be safely disabled for DA RPMs!)

PR #597, opened 2018-02-28
TPC chi2 set in AliESDtrack(AliVTrack*) constructor

fprino:
This commit is aimed at properly setting the TPC chi2 when converting an AOD track into and ESD track using the constructor AliESDtrack(const AliVTrack *track). 

It is a modification in two classes:
1) AliAODTrack to add an helper method to compute the TPC chi2 from the chi2perNDF, by reversing what done in AliAnalysisTaskESDfilter::Chi2perNDF
2) AliESDtrack to set the TPC chi2 in the constructor that creates the track from an AliVTrack


chiarazampolli:
+1

PR #599, opened 2018-03-01
Set impact parameter in conversion from AOD to ESD tracks

fprino:
This commit is aimed at properly setting the track impact parameter when converting an AOD track into and ESD track using the constructor AliESDtrack(const AliVTrack *track).

It is a modification in two classes:

1) AliAODTrack to add the methods to compute the impact parameter and its uncertainty (following advices and suggestions from Ruben)
2) AliESDtrack to call the calculation and to set the impact parameter in case the track passed to the AliESDtrack(const AliVTrack *track) constructor is an AliAODTrack.

### DIFF
--- a/STEER/AOD/AliAODTrack.cxx
+++ b/STEER/AOD/AliAODTrack.cxx
@@ -549,6 +549,62 @@ void AliAODTrack::SetDCA(Double_t d, Double_t z)
 }
 
 //______________________________________________________________________________
+void  AliAODTrack::GetImpactParameters(Float_t &xy,Float_t &z) const
+{
+  // get track impact parameters
+  Float_t p[2], cov[3];
+  GetImpactParameters(p,cov);
+  xy=p[0];
+  z=p[1];
+  return;
+}
+//______________________________________________________________________________
+void  AliAODTrack::GetImpactParameters(Float_t p[2], Float_t cov[3]) const
+{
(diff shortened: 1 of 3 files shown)


chiarazampolli:
+1

chiarazampolli:
Ciao @fprino ,

I don't know if it was already done, but I would open a porting request to track this change for the next tag. I think it is an important one (even though not crucial for data and MC processing). What do you think?

Cheers,

Chiara

fprino:
Hi,

yes, sure I was planning to open a porting request on Jira once I have completed the "package" of changes needed to allow to apply selections on AOD tracks in an equivalent way as what we do on ESD tracks.
For this, I still need to implement and test some modifications in AliESDtrackCuts, which I hope to complete before the next AliRoot tag.
Once also these modifications are in, I will open the Jira ticket with porting request.

Cheers, Francesco

chiarazampolli:
Ciao Francesco,

Thanks! Isn't it worth to open the request already now, in case the new tag is ready before you finish? 

Chiara

fprino:
Hi,

yes, if I don't make it by this evening with the last bunch of changes, I will open the porting request with the two commits done so far.

Cheers, Francesco

chiarazampolli:
Agreed, thanks!
Chiara

PR #600, opened 2018-03-01
In events w/o TPC filter ITS_pureSA tracks as complementary ITS_SA 

shahor02:
In absence of TPC only ITS pureSA tracks are reconstructed, while the ESD->AOD filtering usually accepts only complementary ITS_SA track. This patch detects such events and redefines all cuts requesting compl. ITS_SA tracks to accept ITS_pureSA track. In the end of the event the cuts are restored.


shahor02:
+1

shahor02:
+1

PR #602, opened 2018-03-02
Fix f0 in TDatabasePDG + add resonances in PYTHIA decay table

mpuccio:
Dear @amorsch , @pzhristov , @qgp ,

with this PR I am introducing the modification done by myself and Anders in order to use AliGenParam for the resonances productions. Anders added the relevant resonant states in the PYTHIA decay table.
The PYTHIA decay table itself should be fine ([here](http://personalpages.to.infn.it/~puccio/pylist.txt) the PYLIST with this PR) but we realised that at first the f0 was produced with wrong parameters (see [here](http://personalpages.to.infn.it/~puccio/injectiontest-ko.png)): this was due to how AliGenParam works + wrongly defined f0 in AliPDG (and then in TDatabasePDG).
To fix this I had to comment a line in the `AliPDG::AddParticlesToPdgDataBase()` method with an old definition of f0 (from @amorsch dated 9 years ago).
That f0 definition was due to an old HERWIG version, however now HERWIG has been updated and contains a different definition of the f0 (we checked this morning with @qgp).
I ran some kine only events to check if the new states and some of the old ones are fine and it looks like everything works as expected (see the mass peaks of some injected states [here](http://personalpages.to.infn.it/~puccio/injectiontest-ok.pdf)).

Could you please check that everything is as expected?
Thanks a lot,
Cheers!
Max


dberzano:
+test

amorsch:
+1

chiarazampolli:
+1

chiarazampolli:
PS. I recommented "+1" as I was not sure whether Andreas' comment was taken into account, as the tests were still ongoing...

PR #609, opened 2018-03-05
Fix build on HLT online system with SHUTTLE and all libraries disabled by AliRoot 'oneline' build

davidrohr:


davidrohr:
@dberzano @mkrzewic : Hi Dario, one of your recent changes breaks the build on the HLT, since it needs all modules but also enabels shuttle. This patch disables your reduced build if the variable HLT_ONLINE_MODE variable is set, which we use on the HLT (and which by chance is very similar to the ALIROOT_ONLINE_MODE that you have already.)

dberzano:
I see, thanks for checking. I have explained the rationale in yesterday's CB. I am wondering whether we can find another way rather than adding yet another environment variable? I know I did too, I am not proud of it :-)

So the problem is:

* we need to build on SLC6 because of the SHUTTLE and the DAs
* I don't want any more to prevent people from using C++11 code (#594 was the ring alarm for me)
* HLT clearly uses very modern compilers (I believe)

Is there any other `-D` option passed to CMake we can catch to detect "HLT mode"? Otherwise, I propose we radically change our approach:

* we get rid of all variables (including mine, `ALIROOT_ONLINE_MODE`)
* we detect the C++ standard in use
* we disable the `EVGEN` modules if the C++ standard is not at least C++11

@mkrzewic @davidrohr what do you think?

davidrohr:
Hi @dberzano ,

as I said, this HLT_ONLINE_MODE variable is already used in other places, therefore I just added it here.

Alternatively, HLT sets -DHLT_DOC=ON in the cmake, so we might use that if you prefer it over the env variable. No idea how this is set in the other builds.

Cheers
David

dberzano:
By the way, given our use case (disable unneeded modules where we can't compile them on SLC6) a check like:

```
printf '#include <iostream>\nint main() {}' | g++ -o /dev/null -xc++ -std=c++11 -
```

is sufficient. Some compilers do support `-std=c++11` without fully supporting C++11 but the check is OK with SLC6's default compiler (GCC v4.4.x).

mkrzewic:
Ciao Dario,

On 06/03/18 10:33, Dario Berzano wrote:
>
> I see, thanks for checking. I have explained the rationale in
> yesterday's CB. I am wondering whether we can find another way rather
> than adding yet another environment variable? I know I did too, I am
> not proud of it :-)
>
> So the problem is:
>
>   * we need to build on SLC6 because of the SHUTTLE and the DAs
>   * I don't want any more to prevent people from using C++11 code
>     (#594 <https://github.com/alisw/AliRoot/pull/594> was the ring
>     alarm for me)
>   * HLT clearly uses very modern compilers (I believe)
>
not really, we use the stock gcc 4.8.5
>
> Is there any other |-D| option passed to CMake we can catch to detect
> "HLT mode"? Otherwise, I propose we radically change our approach:
>
>   * we get rid of all variables (including mine, |ALIROOT_ONLINE_MODE|)
>   * we detect the C++ standard in use
>   * we disable the |EVGEN| modules if the C++ standard is not at least
>     C++11
>
> @mkrzewic <https://github.com/mkrzewic> @davidrohr
> <https://github.com/davidrohr> what do you think?
>
actually we use the HLT_ONLINE_MODE variable quite extensively (in
aliroot) to modify the behaviour of some (HLT and offline) classes
online. But in principle we could add a define, in our case not much of
a problem since we have our own build scripts, but it woud still be nice
to be able to switch this mode on in a supported alibuild way.
Mikolaj

dberzano:
Hi @mkrzewic I was referring to CMake's `-D`, not gcc's so they are not defines but simple definitions of CMake variables. Anyway given that the variable as you said is used extensively, I'd say we move on. I am approving it.

dberzano:
+1

mkrzewic:
Hi, I think a new -D option would be better, say -DHLT_ONLINE_MODE

davidrohr:
Hi Dario, just for reference,
I also meant cmake -DHLT_DOC=ON.
Also the C++11 check should be OK since it is supported by GCC 4.8 (It is also enabled on the HLT).
Furthermore, it would also not be a problem to switch to GCC 6 (which is already installed on the HLT and I also have tested in).

In any case, I see you just approved my PR already :|

PR #611, opened 2018-03-09
Bug fix: particle UniqueID was not retrieved before writing stack!

coppedis:
As stated in the commit message, the UniqueID was not retrieved in the loop where particles are written in the stack, but in the case where the data driven spectator fragmentation is applied, this would imply that all the spectators are written in the stack as coming from the target (the last value of the previous loop being used).
This change is therefore important and need to be pulled asap!
Thanks a lot, 
ch/


chiarazampolli:
+1

chiarazampolli:
Ciao Chiara,

I approved your commit. As it is important, I would ask you to open a porting request to track it in the next tag.

Thanks!

Chiara

PR #614, opened 2018-03-14
Merge update of HLT TRD tracking + related changes to run it on the HLT farm

davidrohr:


davidrohr:
@dberzano @brinick : Hi, this patch affects only HLT. There is only a minor fix to the cmake, which was missing a library dependency inside HLT before, which is not seen running inside AliRoot, but it fails on the HLT cluster. Can you please approve this. Cheers, David

shahor02:
+test

dberzano:
Hi @brinick, it is sufficient to merge if one of the authorized persons comments with `+1` without manually doing it.

PR #615, opened 2018-03-14
Ownership for AliOADBContainer

loizides:
Following the discussion about the memory induced by the AliOADBContainer in 
https://alice.its.cern.ch/jira/browse/EMCAL-135
I implemented the possibility to set the owner ship from outside with SetOwner
(there currently is a flag due to the fact that some emcal objects were not properly written,
the default objects are copies of the standard ones, and so one would double delete.
this can be removed in the future).
Also the setowner in the InitFromFile is commented out so that the behavior of the class
should not change. One can then go through the aliroot/aliphysics code and test
the owner ship case by case.

I also added a GetObjectCache function so that one could keep the object and run range and delete the container. This could be used as very simple caching.

Lastly, I  did some general cleanup

Of course we should discuss these changes, before merging.

### DIFF
--- a/STEER/STEERBase/AliOADBContainer.cxx
+++ b/STEER/STEERBase/AliOADBContainer.cxx
@@ -68,7 +68,6 @@ AliOADBContainer::~AliOADBContainer()
   if (fArray)       delete fArray;
   if (fDefaultList) delete fDefaultList;
   if (fPassNames)   delete fPassNames;
-
 }
 
 //______________________________________________________________________________
@@ -96,10 +95,10 @@ AliOADBContainer& AliOADBContainer::operator=(const AliOADBContainer& cont)
     fLowerLimits.Set(fEntries);
     fUpperLimits.Set(fEntries);
     for (Int_t i = 0; i < fEntries; i++) {
-	fLowerLimits[i] = cont.fLowerLimits[i]; 
-	fUpperLimits[i] = cont.fUpperLimits[i];
-	fArray->AddAt(cont.fArray->At(i), i);
-	if (cont.fPassNames) if (cont.fPassNames->At(i)) fPassNames->AddAt(cont.fPassNames->At(i), i);
+      fLowerLimits[i] = cont.fLowerLimits[i]; 
+      fUpperLimits[i] = cont.fUpperLimits[i];
+      fArray->AddAt(cont.fArray->At(i), i);
+      if (cont.fPassNames) if (cont.fPassNames->At(i)) fPassNames->AddAt(cont.fPassNames->At(i), i);
     }
   }
   //
@@ -152,11 +151,10 @@ void AliOADBContainer::RemoveObject(Int_t idx)
 
   //
   // Check that index is inside range 
-  if (idx < 0 || idx >= fEntries) 
-    {
-      AliError(Form("Index out of Range %5d >= %5d", idx, fEntries));
-      return;
-    }
+  if (idx < 0 || idx >= fEntries) {
+    AliError(Form("Index out of Range %5d >= %5d", idx, fEntries));
+    return;
+  }
   //
   // Remove the object
   TObject* obj = fArray->RemoveAt(idx);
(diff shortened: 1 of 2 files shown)


loizides:
I would like to invite Andreas Morsch and Markus Fasel for reviewing the changes, but I do not know how...
I will send them an email.

STEER/STEERBase/AliOADBContainer.cxx
```diff
@@ -225,22 +220,28 @@ void  AliOADBContainer::CleanDefaultList()
...
-  for (Int_t i = 0; i < fEntries; i++) 
-    {
-	if (fPassNames) if (fPassNames->At(i)) if (passName.CompareTo(fPassNames->At(i)->GetName())) continue;
-	if (run >= fLowerLimits[i] && run <= fUpperLimits[i])
-	{
-	  found++;
-	  index = i;
-	}
+  for (Int_t i = 0; i < fEntries; i++) {
+    if (fPassNames) if (fPassNames->At(i)) if (passName.CompareTo(fPassNames->At(i)->GetName())) continue;
```
mfasDa: This line will be interpreted by modern compilers as "Misleading indentation" - should be three separate lines.

STEER/STEERBase/AliOADBContainer.cxx
```diff
@@ -275,6 +276,37 @@ TObject* AliOADBContainer::GetObject(Int_t run, const char* def, TString passNam
...
+  Int_t idx = GetIndexForRun(run, passName);
+  if (idx == -1) idx = GetIndexForRun(run); // try default pass for this run range
+  if (idx == -1) {
+    // no object found, try default
+    obj = fDefaultList->FindObject(def);
+  } else {
+    if (fArray!=0) {
+      obj = fArray->At(idx);
+    } else {
+      obj = GetObjectFromFile(gFile, run, def, passName);
```
mfasDa: What happens if gFile points to a different file?

loizides: To me this seemed some fallback that I do not fully understand, I left the logic as it was. Maybe one should remove it?

STEER/STEERBase/AliOADBContainer.cxx
loizides: Ok that probably was in the code before. I usually expanded such statements, when I saw them.

STEER/STEERBase/AliOADBContainer.h
```diff
@@ -32,22 +44,25 @@ class AliOADBContainer : public TNamed {
   void   RemoveObject(Int_t index);
   void   AddDefaultObject(TObject* obj);
```
mfasDa: I think here it must be made clear, preferably with a doxygen comment, that the function AddDefaultObject must get objects which are unique, which are not handled as run-dependent objects, otherwise it ends in a double-delete. From the current interface this is not necessarily clear to the users as has been seen with several containers. It should be stated also in the comment of the member definition. What can go wrong will go wrong, at the worst moment!

STEER/STEERBase/AliOADBContainer.cxx
```diff
@@ -324,90 +355,94 @@ void AliOADBContainer::WriteToFile(const char* fname) const
...
+  TIter next(cont->GetDefaultList());
+  TObject* obj;
+  while((obj = next())) 
+    fDefaultList->Add(obj);
+
+  //not yet enabled
+  //SetOwner(); 
+
+  delete cont;
+  delete file;
```
mfasDa: TFile::~TFile will close the file at the same time - it must be clear that objects are not any more file-bound. I am particularly thinking about ROOT hists as calibration objects.

mfasDa:
In general it might be helpful to cover the AliOADBContainer with a unit test included in the ctest suite. The unit test should cover
- Write
- Read
- GetObjectByIndex
- GetObjectByRun
For two or three object types (TH1, TF1, TObjArray with TH1+TF1). Possibly this could be extended by deltaVM for the Read operation.

STEER/STEERBase/AliOADBContainer.cxx
loizides: yes it closes it but you will keep the histograms! I tested it. Please also test it to be sure.

PR #622, opened 2018-03-22
add def owner and test macro

loizides:
Implemented proper ownership handling and added test macro.


loizides:
Could this please be approved, so we can go on developing waiting for the tag to appear after easter

chiarazampolli:
+1

PR #627, opened 2018-03-25
Fix: update AliTPCcalibAlignInterpolation::Process argument to AliVEvent

shahor02:
The AliTPCcalibBase::Process(AliESDEvent*) was changed to Process(AliVEvent*) during adaptation of calib framework
to flat containers/HLT, but the derived AliTPCcalibAlignInterpolation::Process was not modified. As a result, the
derived method is not called and the residual trees are not produced.
At the moment fixing just by dynamic_cast from AliVEvent to AliESDEvent, so the task will do nothing on the HLT.


shahor02:
+1

PR #628, opened 2018-03-28
[EMCAL-135] Fix deep copy in pass names

mfasDa:
Since InitFromFile sets ownership over pass name
container (#622) when setting the target object a
deep copy of the pass name has to be performed in
any case, otherwise the container will crash when
accessing the pass name while retrieving the
corresponding object from the container.


mfasDa:
@dberzano @chiarazampolli It is important to have this fix approved as soon as possible and in the next tag, otherwise the TPC PID response will crash when initializing the splines from OADB.

chiarazampolli:
+1

PR #629, opened 2018-03-28
Fix in STEER/STEERBase/AliTRDPIDParams.cxx

luisabergmann:
A constructor had to be added to AliTRDPIDParams.cxx. Otherwise the 'test' object in the GetThresholdParameters() function is not initialized correctly and thus the function does not return the correct threshold parameters for the PID.
Additionally, an output of the threshold parameters was added to the Print() function, to simplify debugging and analysis.


shahor02:
+1

PR #631, opened 2018-03-28
Update LMee cocktail with the rho spectral function from NA60 (Physics Letters B 757 (2016) 437–444)

otonvd:
Update LMee cocktail with the rho spectral function from NA60 (Physics Letters B 757 (2016) 437–444)


rbertens:
+1

PR #632, opened 2018-03-30
 split AliEMCALRecoUtils into AliEMCALRecoUtils and AliEMCALRecoUtilsBase

gconesab:
First step before moving AliEMCALRecoUtils to AliPhysics

### DIFF
--- a/ANALYSIS/ESDfilter/AliAnalysisTaskESDfilter.cxx
+++ b/ANALYSIS/ESDfilter/AliAnalysisTaskESDfilter.cxx
@@ -64,7 +64,7 @@
 #include "AliESDTrdTracklet.h"
 #include "AliAODTrdTrack.h"
 #include "AliAODTrdTracklet.h"
-#include "AliEMCALRecoUtils.h"
+#include "AliEMCALRecoUtilsBase.h"
 #include "AliESDUtils.h"
 
 using std::cout;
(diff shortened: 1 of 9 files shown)


gconesab:
+test

do not approve yet, please review

chiarazampolli:
Ciao Gustavo,

Do I understand correctly that I should not approve? I reviewed the change in AliAnalysisTaskESDfilter.cxx, there it is simple, and ok, the rest is in EMCAL. If I approve for the filtering task, though, I am afraid that everything will be approved.

Cheers,
Chiara

gconesab:
Ciao Chiara,

Yes, let's just approve the testing and on Monday if there are no comments from Emcal we can approve.
Thanks  .... Gustavo

chiarazampolli:
+test

shahor02:
Hi,
To me it looks ok, though since the AliEMCALRecoUtils derives from the AliEMCALRecoUtilsBase, I don't see why do you need to explicitly alias ExtrapolateTrack... methods of AliEMCALRecoUtils to AliEMCALRecoUtilsBase and add extra 3 commits changing AliEMCALRecoUtils methods to AliEMCALRecoUtilsBase.
Cheers,
 Ruben

gconesab:
Hi Ruben,

the idea is to move this AliEMCALRecoUtils class to AliPhysics, thus the need of aliases. The reason for such move is that since this class is heavily used in Analysis a simple bug fix needs a new aliroot tag, which takes some time and it  is not completely necessary. I will profit the next tag to remove this class from aliroot and put it on AliPhysics.

Since we received not more comments, can be this approved? Thanks

shahor02:
Hi Gustavo,
If AliEMCALRecoUtils inherits from AliEMCALRecoUtilsBase, it inherits also all its methods, it does not matter if you leave it in the aliroot or move to aliphysics. So the aliases are redundant. Anyway, they will not hurt, I'll approve the PR
Cheers,
 Ruben

shahor02:
+1

PR #637, opened 2018-04-04
[EMCAL-39, ALIROOT-7708] Check for active TRU

mfasDa:
Implement check of the TRU status based on the status in
the STU region.


mfasDa:
This is a helper function used to check whether a TRU was active in a given run or not. Used to fix a crash reported in ALIROOT-7708, crashing in simulation for a TRU which was dead in the given run due to 0-value for L0 threshold.

mfasDa:
+1

New function, not changing existing logics. Safe to merge!

chiarazampolli:
Hello,

I guess this error is temporary. @dberzano @ktf , can you confirm?

Chiara

ktf:
Looks like the mac builds are currently down. I guess the CERN network outage of today is still affecting that. Merging this.

PR #638, opened 2018-04-04
[WIP][EMCAL-39, ALIROOT-7708] Disable L0 on dead TRUs

mfasDa:
- Adding active status in L0 algorithm: If active is false
  the L0 algorithm is not executed
- Handling active status based on presence of TRU in STU
  region (#637):
  + Raise AliFatal only in case the threshold is 0 for an active
    TRU
  + For inactive TRUs set active status to false in order to prevent
    TRU from running the L0 algorithm

Fixing crash for several runs reported in ALIROOT-7708.


mfasDa:
Work-in-progrss - don't merge!

@gconesab @guernane @loizides Please review! This commit should cure ALIROOT-7708.

gconesab:
Hello, it seems fine to me.
The only minor remark I have that you can ignore is that If I remember well Dario's directives, the doxygen comments on the method should be on the source cxx and not the header?
Maybe increase the classdef

gconesab:
Hi Markus,

thanks for the discussion. I agree that for inline methods there is no need to do this in the source code, in fact I never do and for such kind of simple lines, I even do not put any comment, and rely on the comment attached to the data member declaration. So I take back my suggestion.

Regards ... Gustavo

mfasDa:
+test

mfasDa:
+test

My fault: Since the branch relied on a commit just merged a rebase was necessary for the build first.

gconesab:
Just a Ping to @guernane to have a quick look. v5-09-26 is about to be tagged and it would be good to have. I would anyway approve this if no comment given before tomorrow

guernane:
The commit looks fine to me, it can be merged

PR #644, opened 2018-04-11
Fix STU region in stream operator

mfasDa:
STU region should be displayed in two ways
- hex number for comparison with DCS settings
- bitset in order to easily read of masked regions


mfasDa:
+1

mfasDa:
This fix is needed to check the STU regions of failed p-Pb OCDB entries after reprocessing, described in EMCAL-42.

PR #647, opened 2018-04-13
Add possibility to require ITS match for TRD tracking seeds

martenole:
Only TPC tracks with ITS hits can be considered for HLT TRD tracking without the usage of std::vector::erase() to make it faster. Small typos for the debug output were fixed.


davidrohr:
+1

PR #650, opened 2018-04-18
[EMCAL-172, ALIROOT-7718, ALIROOT-7649] Disable trigger simulation vi…

mfasDa:
…a reco params

Trigger electronics only createed in case it is enabled
in the EMCAL reco params (new field in reco params with
default value true). In case the trigger electronics is a
nullptr (not constructed in the constructor) the trigger
simulation will not be processed.

Off-status for STU is handled via the DCS region: For OFF
STUs the DCS sends  null values for all entries of the
STU configuration. The off-status can therefor be deduced
from a 0 value in the STU region. In this case the active
status of the STU is set to 0, and for inactive STUs the
function L1 processing the L1 trigger simulation will be
skipped.

### DIFF
--- a/EMCAL/EMCALbase/AliEMCALRecParam.cxx
+++ b/EMCAL/EMCALbase/AliEMCALRecParam.cxx
@@ -60,6 +60,7 @@ fUseFALTRO(kTRUE),
 fFitLEDEvents(kFALSE),
 fUseL1Phase(kTRUE),// Run1 setting //raw signal
 fRejectBelowThreshold(0),
+fSimulateTriggerElectronics(true),
 fTrkInITS(kFALSE) // Run1 setting 
 {  
   InitPIDParametersForHighFlux();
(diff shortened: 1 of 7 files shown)


mfasDa:
@guernane, @loizides Please review!

gconesab:
Hi Markus

It think the more appropriate OCDB object is the sim param and not the rec param, can you consider changing it?

Regards ... Gustavo

mfasDa:
Hi @gconesab 

Thanks a lot for your comment!

First reason why I chose the reco params is that the trigger simulation is (for whatever reason) part of the reconstruction chain and not the simulation chain. So it is steered by AliEMCALReconstructor which is usually steered by the reco params. (I am actually surprised EMCAL decided to handle like this because this prevents the simulation of raw data).

The for the sim params are they part of the OCDB? And are they part of the raw OCDB? As this will be a run-dependent setting the OCDB object switching on/off the trigger simulation cannot be part of the simulation OCDB.

For further discussion it is probably better to keep it within the dedicated JIRA ticket (EMCAL-172) in order to not split the discussion to two places.

Cheers

Markus

mfasDa:
The method to switch on/off the trigger simulation via the reco params is now tested with the crashing production in both directions. Since the reco params are not available at constructor nor at init level an Init function similar to InitClusterizer (called InitTriggerElectronics) needed to be introduced, which is called in FillESDTree. Unlike InitClusters a nullptr is a valid status (trigger simulation disabled) so the function cannot use it to determine the init status and therefor will always check whether it has to init the trigger electronics in case it finds a nullptr.

Based on the discussions in EMCAL-39, EMCAL-172 and #651 we need to determine whether we still want to go for this option to allow switching off the trigger simulation via the reco params.

loizides:
Let's go ahead with the PR.

loizides:
+1

PR #651, opened 2018-04-18
fix L0 time array overflow due to null L0 threshold from OCDB EMCAL-39

guernane:
everything is in the title, should fix the old standing EMCAL-39 ticket.
@mfasDa  should be enough to remove the Fatal (to be turned into an Error?) in the TriggerElectroncis constructor, do you think we should include that in the same PR?


loizides:
Hi, I would keep the two patches separate. Good for me to go on. Constantin

loizides:
+1

mfasDa:
Hi @guernane, Hi @loizides,

Yes, error is the appropriate. This is the default log level for grid jobs.

It is OK for now to separate into two patches but in general one should be cautious with the fragmentation into trivial patches because it will become tricky to find certain commits in the history.

Cheers

Markus

loizides:
Ah ok, my reasoning was to perhaps cherry pick this patch in the future, in case we need to backport , etc

PR #653, opened 2018-04-19
[EMCAL-39] Relax severity from Fatal to Error

mfasDa:
As in #651 a protection against the buffer overflow
resulting from 0-thresholds was introduced, the log
severity can be removed from Fatal to Error resulting
that the code will not crash any more in case a 0-
threshold is found.


mfasDa:
@guernane are you fine with this change?

mfasDa:
+test

mfasDa:
+1

Local rec tests passing, commit will be approved

mfasDa:
+1

PR #654, opened 2018-04-20
TRD PID Parameters: adapt response of code when threshold parameters not available

kagarner:
- add output in AliTRDPIDParams::AliTRDPIDCentrality::GetThresholdParameters() that gives error message when threshold parameters are set to default values  (-999) as necessary for LHC17n
- ensure that return value of AliTRDPIDParams::AliTRDPIDCentrality::GetThresholdParameters() is passed over correctly to AliTRDPIDResponseObject::GetThresholdParameters()
- ensure that AliTRDPIDResponse::IdentifiedAsElectron() only returns default kTRUE if no parameters are available


dberzano:
+test

PR #658, opened 2018-04-29
Changes needed to work with Decayers other than AliDecayerPythia like…

amorsch:
… AliDecayerPythia8

For AliDecayerPythia based on FORTRAN there is a known inconsistency between the index returned by GetFirstDaughter(), GetLastDaughter(), GetMother() ... and the actual position in the C++ particle list. That's the reason whe the offset fIncFortran = -1 is needed in this case. This offset was already present in the original version. The new version sets the offset to zero for other decayers like AliDecayerPythia8.


rbertens:
+test Hi @amorsch , are the added print statement in AliGenParam intentional? If yes then we'll merge the PR, but I just wanted to verify. 
Thanks,
Redmer

rbertens:
@amorsch thanks, PR is on its way to being merged

dberzano:
The macOS builder was off during the weekend because of a power cut. macOS test will be run as soon as possible!

PR #660, opened 2018-04-30
Fix: for 285364<=run<285621 produce Error instead of Fatal

shahor02:
in case of detecting different Kr gain table names


shahor02:
+1

PR #661, opened 2018-05-03
Fix C99 warning

ktf:


ktf:
@davidrohr: I think `isfinite` is not part of math.h when using `C++14` and that the correct thing to have is `cmath` and `std::isfinite`, albeit I cannot guarantee that's what you actually wanted if the thing has to compile using nvcc or stuff like that.

davidrohr:
@ktf : I'll test that with standalone, AliRoot GPU and O2 GPUZ build, and then let you know, but might take 2 or 3 days to get to it.

ktf:
no rush.

davidrohr:
+1

PR #664, opened 2018-05-07
[EMCAL-179] Display dec representation of reg mask

mfasDa:
In order to compare to original files the bitwise
representation is not of help as the original values exist
only in decimal represenation. Therefore both decimal
and bitwise representation are displayed.


mfasDa:
+1

PR #666, opened 2018-05-08
Repeat event generation in case event was rejected.

amorsch:
This should solve the problem with events containing un-decayed pi0 leading to a wrong 
Dalitz mee distribution 
(https://alice.its.cern.ch/jira/browse/ALIROOT-7792)


chiarazampolli:
+1

PR #668, opened 2018-05-09
Fix iterator, which fails in GCC 8 static assertion.

davidrohr:
WIthout checking the code in detail: apparaently this iterator should be std::set<int> iterator, instead of <int, int>.
No idea why it worked before, probably by chance.


aphecetche:
+1

PR #667, opened 2018-05-08
Moving STAT-dependent classes from STEERBase to STAT

pzhristov:
This is the first part of the changes - the second one is to make STAT independent from STEERBase.

PR #669, opened 2018-05-09
Add missing STAT in various include_directories

dberzano:
Complements #667 and fixes a problem found with DA RPM builds


dberzano:
+1

PR #678, opened 2018-05-16
Propagate DetectorStatus to AODs

wiechula:
In order to filter for HV dip events in the TPC, the Detector status
flags need to be propagated to AODs.
This also unifies the treatment of the detector flag testing between the
different event types. A virtual function was added in AliVEvnet which
retruns 'true' by default in the detector on check. It is then
overridden by the ESD and AOD implementation


shahor02:
+1

chiarazampolli:
Hello @wiechula , 

Can you please check? It should be a trivial change.

Cheers,

Chiara

wiechula:
Ok this should be fixed.

shahor02:
+1

PR #683, opened 2018-05-22
Fix CA-Glo interplay (ALIROOT-7743)

mpuccio:
Hi @shahor02,

this should fix the issue we encounter with the upgrade production.
It was due to the change of ordering in the cluster and the change of frame.
This should address both the issues.

Cheers,
Max


shahor02:
+1

Thanks Max!

PR #684, opened 2018-05-23
Improve install speed on macOS (see O2)

mpuccio:
Ciao @dberzano ,

this is on the same line of what you did for o2. It does need a scrutiny before merging.

Cheers,
Max


dberzano:
+1

PR #685, opened 2018-05-24
Fixes needed for compilation with gfortran8

mfasDa:
Compilation with gfortran8 failing due to
mismatch in array dimension when calling
subroutines. Array dimensions were adapted
where needed.


mfasDa:
There were two more files to be adapted. After this AliRoot compiles.

dberzano:
+test (@qgp any objections?)

mfasDa:
Some more details:

gfortran8 is raising a compiler error if a field in a subroutine has as larger! dimension as the field the subroutine is called with. As homebrew updated gcc to version 8 aliroot will stop compiling on mac due to these errors.

What I did to circumvent this is to initialize the fields when needed with the same size of the field in the subroutine. As far as I understand this should not have any affect on the program behavior as the additional values were undefined in any case. For data blocks I initialized the additional values with 0. Also here these values were anyhow undefined so the program behavior should not change. But I didn't have the chance to test this by i.e. running a simulation.

dberzano:
+1 @mfasDa thanks for your efforts!

PR #686, opened 2018-05-26
Treat AliAODVertex::GetNContributors like in ESDVertex

shahor02:
The AODVertex made from primary VertexTracks will get NContributors assigned from the ESDvertex.
Old behaviour of counting stored primary daughters may be enforced using static method
SetUseCountVtxTrackContributors(kTRUE); // off by default
For old AODs, where fNContributors was not stored, GetNContributors will fall back to old way of counting.


shahor02:
+1

PR #689, opened 2018-05-29
Speed up installation only outside Online Mode

dberzano:
Generation of DA RPMs is broken because of #684, this one fixes it


dberzano:
@mpuccio fyi

PR #691, opened 2018-05-31
bug in determination of fIncFortran corrected

amorsch:
affects
v5-08-13zl-cookdedx-1
v5-08-13zk-cookdedx-1
v5-08-13q-p13-01-1 
v5-08-13q-p12-01-1
v5-08-13q-p13-01-cookdedx-1
v5-08-13q-p13-01-cookdedx-1
but can only have an effect on MC production  in case of injected signal production with AliGenParam.


rbertens:
+1

chiarazampolli:
Hello,

I think that also v5-09-30 is affected, as it contained https://github.com/alisw/AliRoot/pull/666, which was the origin of this issue, right?

Chiara

PR #692, opened 2018-06-04
[EMCAL-212] Fix FPE exception

aiola:
This PR should fix the issue reported in EMCAL-212.
I believe it is caused by events that have no background patches. This should be quite rare in PbPb collisions, which is probably why it was not spotted before.


shahor02:
+test

shahor02:
+1

gconesab:
+1

PR #693, opened 2018-06-06
Fix trivial bug introduced in #683

mpuccio:
Hi @shahor02,

I introduced a trivial bug with the previous PR, in principle now the cluster sorting will be performed once for both CA and Global Tracker (as it was intended to be in the first PR, but the return was before the sorting of the clusters).

Can you please check if this works for you?
Cheers,
Max


shahor02:
+test

Hi Max, will test now with the same data sample.
Cheers,
 Ruben

shahor02:
+1

shahor02:
Hi @mpuccio 
Yes, it works, thanks! But there is a problem with the PR approval: whether I do +test or +1 it just sends another comment suggesting to do the same. @dberzano , could you please check?
Cheers,
 Ruben

shahor02:
+1

PR #698, opened 2018-06-11
ATO-396  - AliNDFormulaBrowser changes

miranov25:
# ATO-396 - Improvement for formula browser ￼…
- browser not blocking main window
- bug fix in windows resize
- reformat code  using automatic intendation

# ATO-46 - fix AliExternalInfo and AliTreePlayer tests for root6
* test running in ROOT5 did not work in ROOT6
* Compiled macro can not be executed  more than one in ROOT 6 ?
* Way aroud - discableing compiled macros


miranov25:
+1

PR #702, opened 2018-06-13
Store new Beta* leveling records in LHCData

shahor02:
Store new records in LHCData:
ALI_Lumi_Source_Name
LHC_BetaStar_TargetOptic
LHC_BetaStar_Active


shahor02:
+1

PR #707, opened 2018-06-19
ATO-445 - Interface for NDimensional function representation TMVA and THn)

miranov25:
ATO-445 - Interface for NDimensional function representation TMVA and THn

### AliNDFunctionInterface
static implementation enable usage of EvalMVA in the TFormula and TTreeFormula
* variadic function to create vector and evaluate regression
* Interface for the THn interpolation
  * linear interpolation of the THn content
* Interface for the TMVA based regression representation
  * FitMVA
  * LoadMVAReader
  * EvalMVA

### QAtrendingfitExample.C
Example code using AliNDFunctionInterface for TMVA regression of QA data

###  To notice
* Code is still experimental. User interface can change
* C++11 features are used here. For the moment only compiled macro
* For the moment only root6 TMVA interface supported
  * To check if root5 can be also supported


miranov25:
+1

PR #711, opened 2018-06-26
Fix memory leaks in DigitizeWithTailAndCrossTalk

mario6829:
Memory leaks in method DigitizeWithTailAndCrossTalk of AliTPCDigitizer were found and fixed.

### DIFF
--- a/TPC/TPCsim/AliTPCDigitizer.cxx
+++ b/TPC/TPCsim/AliTPCDigitizer.cxx
@@ -799,6 +799,9 @@ void AliTPCDigitizer::DigitizeWithTailAndCrossTalk(Option_t* option)
     for (Int_t ires = 0;ires<20;ires++) timeResArr->AddAt(graphRes[ires],ires);
     timeResFunc.AddAt(timeResArr,isec); // Fill all trfs into a single TObjArray 
     nIonTailBins = graphRes[3]->GetN();
+    delete [] graphRes;
+    delete trfIndexArr;
+    delete timeResArr;
   }
 
   //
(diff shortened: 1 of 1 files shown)


miranov25:
Hallo Mario.

Thank you for your modification. I checked the diff and unfortunately I can not approve it.
I think your test did not fail because you did not enable common mode correction.
Objects have to be deleted at the end of the digitizer.

Marian

mario6829:
Hello Marian,
thanks for suggestions, I'll modified the code and check again. How can I activate the common mode corrections ?
Thanks. Cheers
Mario

MarekKowalski1504:
+1

wiechula:
Hi @MarekKowalski1504, @miranov25  wrote that this is not yet ready for merging...

miranov25:
Hallo Mario.

We should use special witch in MC:
dpgsim.sh --mode full --run 246751 --generator Custom --nevents 1 --seed 3212 --ocdb cvmfs --simulation SimulationDefaultIonTail
+ 
change in OCDB.  This is difficult to explain as we are usually patchin OCDB snapshots.
In case you can use any anchor production you can use snapshot as in :+1: 
 https://alice.its.cern.ch/jira/browse/ALIROOT-7077

But I assume once you put destruction at the end of the function - there will be no problem.

Marian.

MarekKowalski1504:
I have accepted it for merging... See my comment above.

miranov25:
Hallo.
We should revert the merge and correct the code. As I wrote 3 hours ago, I assume the code will fail
in case of simulating ion tail. In test of Mario he did not check this scenario

Marian

mario6829:
I moved the delete to the end of the method. I am re-running valgrind just to be sure there are still no leaks (since I was not able to move all 'delete' - I'll report with more details when I submit the code). I suppose I need a new PR, since this one is now closed...

PR #712, opened 2018-06-26
New correct fix for memory leaks in AliTPCDigitizer

mario6829:
Previous fix (#711) was not correct, since it failed when simulating common mode corrections. This better fix should solve the memory leaks without introducing any flaw.
Essentially, all 'delete' were moved to the end of the method (the TOBjArray itself is not explicitly deleted since doing so causes a crash). Valgrind tests prove the absence of memory leaks.


MarekKowalski1504:
+1

PR #715, opened 2018-06-28
TimeStamp precise up to the BC

shahor02:
The GetAliTimeStamp and GetTimeStampCTPBCCorr (changed to double) will return the time-stamp
precise up to the BC. Note that rarely the 1st CTP scaled is not in sync with the actual CTP start,
in this case the CTP time stamp may be wrong. If it cannot be calculated, the GetTimeStamp (event building time)
will be assigned and the AliTimeStamp (from GetAliTimeStamp) will have GetUniqueID() == 1 (in case of success it is 0)


shahor02:
+1

PR #717, opened 2018-06-29
Added method AliLumiTools::GetLumiFromDIPInst

shahor02:
Will extract from the LHCData inst. lumi with high frequency sampling


shahor02:
+1

PR #720, opened 2018-07-03
Rename GPU CMAKE build options

davidrohr:
Rename cmake option as discussed with Max and Dario (use same common name for AliRoot / O2; TPC / ITS)

### DIFF
--- a/HLT/CMakeLists.txt
+++ b/HLT/CMakeLists.txt
@@ -13,8 +13,8 @@
 # * provided "as is" without express or implied warranty.                  *
 # **************************************************************************
 
-option(HLT_CUDA   "Build HLT GPU tracker using CUDA"   OFF)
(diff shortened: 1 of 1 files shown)


dberzano:
+test @davidrohr does this have an impact on the infamous "online" builds where we need to stick to C++98?

davidrohr:
@dberzano : In AliRoot this flag is only used when built for the HLT farm. Doesn't change anything for the rest.

I just renamed it as well to be consistent for my own convenience in AliTPCCommon.

dberzano:
@davidrohr I am afraid we are using an outdated CMake version for our Run 2 builds. The problem has to be fixed on my side. Once fixed, it should pass the test. I am keeping track of a number of changes here: alisw/alidist#1214

PR #721, opened 2018-07-05
ATO-418 -AliDrawStyle (CSS) and AliPainter

miranov25:
### AliPainter:
    1. Refactored;
    2. Changes in interface;
      a. Now the object it's first argument in function (in previous version it was string);
      b. DrawHistogram is attempt to implement THnBase::Draw()
    3. New tests added;
    4. Limits could be specify with using statistical values;
    5. Support of expressions in limits;
    6. THn was changed to THnBase;
    7. Bugs fixed, stability improved;
###   AliDrawStyle:
    1. "margin" property was added (common property for margin-(top,left,right,bottom);
    2. GetNamedTypeAt and PrepareValue was changed to Float_t from template;
    3. If index will be not find, will be use last index in array;
    4. In doxygen used generator of images;
    5. AliDrawStyle based on --std=c++98;
    6. Examples figTemplate*.css was changed according with new syntax;

###   Documentation for AliDrawStyle and AliPainter extended.
  In doxygen used generator of images.
  Examples in documentations was fixed according with new syntax;
  Code was installed without error and test passed.

### DIFF
--- a/STAT/AliPainter.cxx
+++ b/STAT/AliPainter.cxx
@@ -13,12 +13,6 @@
  * provided "as is" without express or implied warranty.                  *
  **************************************************************************/
 
-
-/*
-* TODO: add checks for nonexistent projections @Boris
-* TODO: add auto canvas->cd(1) after DivideTPad
-*/
-
 #include "AliPainter.h"
 #include "AliTMinuitToolkit.h"
 #include "TPad.h"
@@ -28,6 +22,7 @@
 #include "TMultiGraph.h"
 #include "TGraph.h"
 #include "TError.h"
+#include "THnBase.h"
 #include "THn.h"
 #include "TH1.h"
 #include "TH2.h"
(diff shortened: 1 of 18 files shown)


miranov25:
+1

bdrum:
Hello,

how can I reproduce the problem? At lxplus that class was compiled without errors.

STAT/AliPainter.cxx
```diff
@@ -573,241 +420,577 @@ std::map<TString, Double_t> AliPainter::statValues;
...
+
+    if (match == 0 && isChange) {
+      atts.push_back(TString(exprsn(startIndex + 1, finishIndex - startIndex - 1)));
+      isChange = kFALSE;
+    }
+  }
+
+  if (verbose == 4) {
+    TString infoString = "";
+    for (auto const& value: atts)
```
dberzano: This is a C++11 construct as the error message says, and it's not allowed since we - quite unfortunately - need to have this code work with the native SLC6 compiler. Only C++98 is possible here.

bdrum: Ah, ok, thanks!

miranov25:
+test

miranov25:
+1

PR #723, opened 2018-07-06
Fix deletion order issue in SSD geometry

marcovanleeuwen:
This fixes an issue with deletion order in the SSD geometry: the objects in a TList were deleted before the TList was deleted; this causes a problem because the TList accesses the objects to read a deletion flag when the TList itself is deleted. This was mostly a formal problem, but best to fix it anyways.


dberzano:
+1 do you know if this is related to the MC crashes at the end that we see for instance in [ALIROOT-7894](https://alice.its.cern.ch/jira/browse/ALIROOT-7894)?

PR #724, opened 2018-07-07
ATO-418: AliPainter and AliDrawStyle - doxygen changes 

miranov25:
### ATO-418: AliPainter and AliDrawStyle 

￼Images for documentation are generate in when test script called.
Small bug in AliPainter fixed. (ParseString always creates first element as empty in output vector).
More examples of usage parsing funtions was added.
Fixed and test all examples in doxygen documentation.


miranov25:
+1

PR #727, opened 2018-07-13
ATO-445 - Bootstrap emulation + regression error estimates

miranov25:
* adding support for vectors of MVA regression

### AliNDFunctionInterface
* extended Doxygen documentation
* adding global verbosity flag
* ADD:
  * map of registered method array
  * LoadMVAArrayReader - to load MVA method arrays
  * EvalMVAStatArray - to evaluate statistic on top of regression array


miranov25:
+1

PR #728, opened 2018-07-13
New B and Lambda_b decays added for HF upgrade production

btrzecia:
Create a new case: kHFYellowReport, for HF upgrade production with new B0  and Lambda_b decay channels forced.

### DIFF
--- a/EVGEN/AliDecayer.h
+++ b/EVGEN/AliDecayer.h
@@ -20,9 +20,9 @@ typedef enum
     kAll, kNoDecayHeavy, kHardMuons, kBJpsi,  kBJpsiUndecayed,
     kWToMuon,kWToCharm, kWToCharmToMuon, kZDiMuon, kZDiElectron, kNeutralPion, kAllMuonic,
     kChiToJpsiGammaToMuonMuon, kChiToJpsiGammaToElectronElectron, kNoDecayBeauty, kPsiPrimeJpsiDiElectron,
-  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,kHFYellowReport,
+  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,
   kHadronicDWithV0,kHadronicDWithout4BodiesWithV0,kAllEM,
-    kLcpKpi, kLcpK0S
+    kLcpKpi, kLcpK0S, kHFYellowReport
 } Decay_t;
 #endif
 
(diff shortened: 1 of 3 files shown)


dberzano:
+test

EVGEN/AliDecayer.h
```diff
@@ -20,7 +20,8 @@ typedef enum
     kAll, kNoDecayHeavy, kHardMuons, kBJpsi,  kBJpsiUndecayed,
     kWToMuon,kWToCharm, kWToCharmToMuon, kZDiMuon, kZDiElectron, kNeutralPion, kAllMuonic,
     kChiToJpsiGammaToMuonMuon, kChiToJpsiGammaToElectronElectron, kNoDecayBeauty, kPsiPrimeJpsiDiElectron,
-  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,kHadronicDWithV0,kHadronicDWithout4BodiesWithV0,kAllEM,
+  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,kHFYellowReport,
```
rbertens: Hi, trivial question maybe, but here kHFYellowReport is added in between of the list of elements of Decay_t . Are we sure that at other points in the code, Decay_t isn't compared to integer values ?

dberzano: This is a very good point. It must be added at the end of the enum definition in any case. @btrzecia please update the PR as requested by @rbertens.

EVGEN/AliDecayer.h
```diff
@@ -20,9 +20,9 @@ typedef enum
     kAll, kNoDecayHeavy, kHardMuons, kBJpsi,  kBJpsiUndecayed,
     kWToMuon,kWToCharm, kWToCharmToMuon, kZDiMuon, kZDiElectron, kNeutralPion, kAllMuonic,
     kChiToJpsiGammaToMuonMuon, kChiToJpsiGammaToElectronElectron, kNoDecayBeauty, kPsiPrimeJpsiDiElectron,
-  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,kHFYellowReport,
+  kElectronEM, kGammaEM, kDiElectronEM, kBeautyUpgrade,
```
dberzano: Could you just fix the alignment here?

EVGEN/AliDecayer.h
btrzecia: Hi, indeed that was not a safe solution, thanks for pointing this out.

rbertens:
+1 I realize that this PR was never merged. From my side it's OK, but it would need also approval of the 2nd list of approvers.

chiarazampolli:
+1

PR #729, opened 2018-07-14
ATO-418, ATO-445 - Adding tutorials + AliPainter,CSS updates  + Doxygen

miranov25:
# ATO-418: AliPainter and AliDrawStyle (Boris)
##  AliPainter
* Warning output for PandasParser was simplified.
##  AliDrawStyle:
* Bug with margins(if specified side-margin it was applied to all margin) fixed;
* Added ndivision property(ndivisions for both -x,y and x-ndivisions, y-ndivisions);
* Function ReadCSSString was added. Now you can copy content of your css file to TString and parse it to array;
* List of available properties was added to the ApplyCssStyle description;
Code was installed without error. All tests passed.

 
# Tutorials - Create Notebooks directory  (Marian)
* TPCDataVolume.C macro with parallel  TPCDataVolume.ipynb
* Add Notebooks to test and tutorials
* add test to default tests
* extend doxygen documentation - including generatig figures
  * including log output from macro does not work
* check with root team setup for the macro
````
///\macro_output
///\macro_code

PR #730, opened 2018-07-18
ATO-418, ATO-445 - Adding tutorials + AliPainter,CSS updates + Doxygen

bdrum:
Hello Dario,

I also change our code according to your review:
https://github.com/alisw/AliRoot/pull/729#pullrequestreview-137778077

I'm not sure about new pull request, but we can't use your modifications, because 
we need to work with such strings like "$AliRoot_SRC/STAT/test/..." all our tests and examples depends on it, so I added hard coding replacing of such variables, may be we can do this modification now, but later we can change to more nice solution.

I expect that with your commit AliDrawStyleTest will be failed.

Boris


dberzano:
I am seeing some confusion here. I had _temporarily_ closed #729. The idea is that we _reopened_ it instead of creating a new one. However, commits mismatch between this PR and the other. Let's continue the discussion on the _other_ PR please.

PR #731, opened 2018-07-18
Fix: AODtracks added via ConvertTRD should have fDetPid filled

shahor02:
Details in https://alice.its.cern.ch/jira/browse/ALIROOT-7874?focusedCommentId=217574&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-217574

Also, AliAODTrack methods like GetTPCmomentum() which extract the info from the fDetPid will fall
back to the track momentum in absence of the fDetPid (like it is done in the AliESDtrack)


shahor02:
+1

PR #733, opened 2018-07-23
 ATO-432 - Adding optional stream to DUMP HLT clusters to the tree 

miranov25:
### ATO-432, ATO-436- temporary protection against division be 0 
-  proper fix checking condition  in TODO

## ATO-432 - Adding optional stream to DUMP HLT clusters to the tree 
*  kStreamClDumpLocal - to dump clusters
*  to be used for high dedx particle trigerring and MC tunning
  * fragmentation (calibration, trip studies)
  * magnetic monopoles
  * adding global ID to the stream


miranov25:
+1

PR #737, opened 2018-08-01
[EMCAL-32] Added optimized clusterizer v3

rhaake:
This is another implementation of the algorithm in AliEMCALClusterizerv2.
Several optimizations have been applied, the clustering is now done recursively around the seeds.

Local tests show an identical clustering behavior as for clusterizerv3. More details in the JIRA ticket [EMCAL-32].


EMCAL/EMCALrec/AliEMCALClusterizerv3.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include "AliEMCALClusterizerv1.h"
+class AliEMCALRecPoint; 
+class AliEMCALDigit;
+
+// Define numbers rows/columns for topological representation of cells
+const Int_t kNrows    = (24+1)*(6+4);  // 10x supermodule rows (6 for EMCAL, 4 for DCAL). +1 accounts for topological gap between two supermodules
+const Int_t kNcolumns = 48*2+1;        // 2x  supermodule columns + 1 empty space in between for DCAL (not used for EMCAL)
+
+struct cellWithE {
```
mfasDa: Sorry, I overlooked this in the first round: Since the struct is only used in AliEMCALClusterFinder I would make this a private nested struct of the AliEMCALClusterFinder and not expose it to the public.

EMCAL/EMCALrec/AliEMCALClusterizerv3.h
```diff
@@ -0,0 +1,122 @@
...
+    TObjArray*          GetFoundClusters() {return fFoundClusters;}
+    
+  private:
+    AliEMCALRecPoint*   GetClusterFromNeighbours(AliEMCALRecPoint* recPoint, Int_t row, Int_t column);
+    void                GetTopologicalRowColumn(AliEMCALDigit* digit, Int_t& row, Int_t& column);
+
+    AliEMCALGeometry*   fEMCALGeometry;
+    cellWithE           fSeedList[kNrows*kNcolumns];
+    AliEMCALDigit*      fDigitMap[kNrows][kNcolumns];
+    Bool_t              fCellMask[kNrows][kNcolumns];
```
mfasDa: The three arrays are working arrays which I think should be transient (adding "//!<!")

gconesab:
+test

EMCAL/EMCALrec/AliEMCALClusterizerv3.h
```diff
@@ -0,0 +1,122 @@
...
+ * See cxx source for full Copyright notice                               */
+
+
+#include "AliEMCALClusterizerv1.h"
+class AliEMCALRecPoint; 
+class AliEMCALDigit;
+
+// Define numbers rows/columns for topological representation of cells
+const Int_t kNrows    = (24+1)*(6+4);  // 10x supermodule rows (6 for EMCAL, 4 for DCAL). +1 accounts for topological gap between two supermodules
+const Int_t kNcolumns = 48*2+1;        // 2x  supermodule columns + 1 empty space in between for DCAL (not used for EMCAL)
```
mfasDa: The segmentation in phi (rows) is strictly speaking not correct because the small sectors only have 1/3 of the number of rows. Like this you allocate more than you actually need. And for run1 there is no DCAL which even further increases the unused memory. Also the gaps correspond to memory which is actually not used and maybe one can do the supermodule separation algorithmically.

Ideally the dimensions are taken from the AliEMCALGeometry, however in this case the arrays would have to be allocated dynamically.

EMCAL/EMCALrec/AliEMCALClusterizerv3.cxx
```diff
@@ -0,0 +1,299 @@
...
+  // --> Recursive to neighboughs and create cluster
+  // --> Seed cell and all neighbours belonging to cluster will be put in 2D bitmap
+
+  // Reset digit maps and cell masks
+  for(Int_t row=0;row<kNrows;row++)
+    for(Int_t column=0;column<kNcolumns;column++)
+    {
+      fCellMask[row][column] = kFALSE;
+      fDigitMap[row][column] = 0;
+    }
```
mfasDa: Can be done with memset (faster)

shahor02:
+test

shahor02:
Hi @rhaake , as far as I understood from the jira, this will not be used in productions at the moment, right?

rhaake:
Hi Ruben,
Exactly, for the moment the idea is to have the implementation in the AliRoot to allow large-scale tests.
Best,
Ruediger

EMCAL/EMCALrec/AliEMCALClusterizerv3.h
rhaake: Thanks, this is true. I have also put the const definitions into a namespace.

EMCAL/EMCALrec/AliEMCALClusterizerv3.h
rhaake: Yes, and it seems I forget to give a basic documentation for the class member. Added.

EMCAL/EMCALrec/AliEMCALClusterizerv3.h
rhaake: I think here we have to trade-off between speed and memory. The gaps are part of the algorithm by setting up a topology.

We should probably have in mind that the overall memory consumption of the topological maps is  (24+1) rows × (48×2 +1) columns ×(6+4) supermodules ×(1 Byte from cellmask+ 8 Bytes digitmap + (4+4+4) Bytes struct array) = 510 KB

I think we could as you say save memory, but on the cost of additional computations.
Getting the topological row/column while taking into account the small sectors requires at least "one subtraction operation" more per cell as far as I can see.
This could save us 4* (8 rows * 97 columns) * (1 Byte from cellmask+ 8 Bytes digitmap + (4+4+4) Bytes struct array)
= 65 KB

Leaving out the one-cell gaps would probably require many more computation operations. We would have to check the SM location algorithmically for each cell. This is naturally given with the one-cell gaps.
And here we are even talking about less memory to save:
(10 rows * 96 columns) + (250 rows * 1 column) * (1 Byte from cellmask+ 8 Bytes digitmap + (4+4+4) Bytes struct array) = 25 KB

So I think if we are not strongly memory-limited, the current approach is to be preferred in my opinion.

EMCAL/EMCALrec/AliEMCALClusterizerv3.cxx
rhaake: Thanks, I do it with memset.

rhaake:
The suggested comments have been implemented. Hopefully, I didn't introduce new problems.
Can I trigger a +test without merge permissions?

mfasDa:
Hi @rhaake,

Thanks for the fixes! No further comments from my side - I think it is ready for merge.

No, you can't, even EMCAL approvers seem to not have approval rights even though code is fully within the EMCAL directory (??????), must be approved by the offline experts.

Cheers

Markus

dberzano:
+1 Whenever a `CMakeLists.txt` changes, we need to give the approval. This is deliberate.

PR #740, opened 2018-08-01
Fix for ROOT 6 hardcoded paths error

dberzano:
We introduce a new `generate_dictionary_flat` CMake function to be used when
header files are in a subdirectory. The new function strips paths when passing
them to `rootcint`, hardcoding just the base header names into the libraries.
At runtime, when neither the source and the build directories are available,
basenames will be searched into the ROOT_INCLUDE_PATH environment variable and
since we install our header files in a flat directory it will finally work.

Note that the problem we are addressing does not show up in local installations
as ROOT hardcodes the full original path first, so the header files will be
always (wrongly) found in the source directory.

With @davidrohr.


dberzano:
+1

PR #741, opened 2018-08-02
[EMCAL-214] Adding range check for invalid digit coordinate

mfasDa:
As discussed in the ticket digits with coordinates out-
of-range will be ignored as they are coming from
corrupted data. Instead an HLTError will be raised.

For the range check the internal boundary check of
function GetAbsCellIdFromCellIndexes is used and its
return value -1 is handled in the function AddDigit.


mfasDa:
@davidrohr: The function GetAbsCellIdFromCellIndexes in AliEMCALGeometry has an internal boundary check performed at the beginning of the function that always returns -1 in case the cell position is outside limits in both directions (in supermodule, eta and phi). In this case the cell ID is assured to be within limits in case it is not -1, but the value -1 stands for any error state of the function. With catching the case -1 I can just reuse the boundary check applied before. Unfortunately error handling and return handling are mixed up and not separated in this development.

davidrohr:
+1

PR #746, opened 2018-08-07
[EMCAL-369] Speedup of raw fitter for HLT

mfasDa:
Two steps found to speed up the EMCAL raw analyzer / raw fitter

- Replace in-loop dynamic allocations of memory in raw fitter by class member objects 
- Cache ALTRO mapping in raw analyzer component and pass it over to the calo raw stream in order to have the expensive read operation of the ALTRO mapping from disk only once

Now profilers show that the performance of the raw analyzer is dominated by the least square fit.


mfasDa:
+1

For the EMCAL part.

davidrohr:
+test

davidrohr:
+1 Improved EMCAL HLT code, running in production already

PR #747, opened 2018-08-13
Exporting base classes and functions for TPC space-charge studies to AliTPCCommon repository

ehellbar:
Aliroot independent classes used in AliTPCSpaceCharge3DDriftLine have been moved from TPCbase to external AliTPCCommon repository. Modified AliTPCSpaceCharge3DDriftLine to be used as interface class. Added building of TPCSpaceChargeBase library to TPCbase cmake.
This commit does not affect any classes used in official production chains.


miranov25:
Hallo Ernst.

If I understand correctly, new AliRoot after update, will be not compiled without updating AliTPCCommon.
In such case we should approve pull request in the same time, and announce to alice-offine that they 
should rebuild AliRoot. 
Is the modification in AliTPCCommon already in?

Marian

ehellbar:
Hello Marian,
the latest AliRoot release v5-09-35 is built against AliTPCCommon v1.3.1. All the changes are already implemented in this AliTPCCommon tag. However, there are some string literal warnings when using Form(), which can be treated as errors, depending in compiler options. Those are fixed in the next AliTPCCommon tag.
So I think you're right, it's better to wait for the next AliRoot release, which will be built against AliTPCCommon v1.3.2.
Cheers,
Ernst

ehellbar:
Hello Marian, 
since yesterday the new AliRoot release is built with AliTPCCommon v1.3.2. All warnings are fixed in this tag, so we can proceed with the PR.
Cheers,
Ernst

wiechula:
+1

dberzano:
+1 in principle we can spot such problems during the build. I confirm AliRoot v5-09-36 is built with AliTPCCommon v1.3.2 (we even have v1.3.3 ready, next AliRoot version will come with it).

PR #749, opened 2018-08-22
Add run limits for 2018 (from Marie)

fprino:
This commit implements a modification from Marie Germain to set the run number limits for 2017 and 2018.
It is needed to have proper results forthe automatic QA scripts for MCs anchored to 2018 data.


chiarazampolli:
+1

dberzano:
Mmmmmhhh... even if for now this is OK-ish, I strongly believe we should get rid of this. On AliEn and CVMFS we have already this mapping in the OCDB:

```
/cvmfs/alice-ocdb.cern.ch/calibration/data/OCDBFoldervsRunRange.xml
```

This file is stored on AliEn and automatically synced to CVMFS. As for the run type, it can be deduced from elsewhere I believe.

chiarazampolli:
Ciao Dario,
Yes, we know, but for now we decided to "fix" it this way, to allow the QA to be processed. Meanwhile we will try to see if a better way can be found.
Chiara

PR #750, opened 2018-08-24
Change fortran comment to allow compilation with ninja

dstocco:
This is a preparation commit to be able in the future to compile with ninja.
The main problem is that there are /* in 3 comment lines in the fortran file that are interpreted as C comments by ninja.
This commit therefore only modifies commented lines in the code, which means that has no impact whatsoever.
@dberzano: can you please check it?


rbertens:
+test

PR #751, opened 2018-08-26
Fix for T0 MC-embedding

shahor02:
In case of multiple inputs T0 was skipping all inputs but last one (background),
creating digits in the AliT0 copy associated with its loader and then using for
TreeD the empty digit container of the AliT0 loaded from the 1st input. Now all
inputs are incrementally processed and digits are registered in the AliT0 of the
1st input


dberzano:
+1

PR #752, opened 2018-08-27
added extra getters to AliTPCTransform, they are needed for TPCFastTransformation

sgorbuno:


davidrohr:
+1 This is only a preliminary PR adding some Getters to AliRoot that are needed for the next Version of AliTPCCommon. The real PR that will adopt the new transformation will come later.

PR #754, opened 2018-08-28
[EMCAL-402] Cache ALTRO mapping in the TRU raw analyzer

mfasDa:
The ALTRO raw stream is constructed for every event and
therefore always reads the same ALTRO mapping from a text
file from disk. As the altro mapping doesn't change it is
can be cached in memory and passed to the raw analyzer
from cache, which is a significantly faster operation than
reading it from disk.


davidrohr:
+test DO NOT MERGE YET, needs to be tested in HLT cluster first.

davidrohr:
+1 tested in HLT with data replay, seems stable, now enabled in production in the cluster, approved

PR #756, opened 2018-08-29
Work around TGeoHMatrix bug in ROOT v6-14-04

dberzano:
`operator=()` does not work properly any longer if a null pointer is passed as
value. The intended behaviour is to invoke `Clear()` to the given instance of
`TGeoHMatrix`. This fix in AliRoot makes it work as intended without waiting for
a fix in ROOT.


dberzano:
+1

I am approving this as it is required for the ROOT 6 release validation, but it's better in any case if the TOF experts could take a close look. Worst case, we revert.

PR #757, opened 2018-08-30
Fix processing with ROOT6

davidrohr:
Move rpc.h to .cxx, since on the GRID we might compile on a
different OS than we run, and ROOT will face compatibiltiy
issues parsing the header otherwise.


davidrohr:
+test

PR #762, opened 2018-09-01
Disable CheckTasks() in ROOT 6

dberzano:
Since the introduction of automatic library loading and rootmaps, this
functionality should not be needed any longer, see ALIROOT-7936


dberzano:
+1 @davidrohr (if anyone sees a problem here, please speak up!)

PR #765, opened 2018-09-05
Add missing TF1 streamers for ROOT 6 compatibility

hcab14:
This is needed for AD/Calib/Saturation OCDB objects where TF1 streamers are missing which prevents correct loading of such files in ROOT6.


dberzano:
+1 Thanks! I am surprised that the solution is so simple!

PR #768, opened 2018-09-05
[WIP] fix gROOT->ProcessLine("AliAnalysisTaskSE::Class()") ROOT6 problem properly

davidrohr:
This works around the gROOT->ProcessLine command, by registering a static callback function from AliAnalysisTaskSE into AliAnalysisManager, which allows to obtain the class pointer without macro invocation and without cyclic library dependencies.

PR #769, opened 2018-09-07
Workaround avoiding gROOT->ProcessLine() in a simpler way

davidrohr:
Simpler version instead of alisw/AliRoot#768

@dberzano : How about this check, which is actually much simpler (but slower as it does string compares).


davidrohr:
+test

dberzano:
I like it :-)

String comparison is slower, but the check function does not even appear in profiling, it's executed only once at the very beginning, so it's OK!

Tomorrow I'll test it and merge it if it works (I think it does).

dberzano:
+1 yes, it seems to be working, thanks! Closing #768 as well.

PR #770, opened 2018-09-10
Add persistent counter for N ESDTracks

shahor02:
Allows to extract number of original ESD tracks when dealing with filtered embedded event


shahor02:
+1

PR #771, opened 2018-09-11
memory leak in AliDCSMessage fixed

hcab14:


hcab14:
by reading more code I realized that in
`  TMap* AliDCSClient::GetValues(...)`
there is this code:
  ```
TObjArray* resultSet = new TObjArray();
  resultSet->SetOwner(1);
  //...
  fResultErrorCode = ReceiveValueSet(resultSet, ownerIndex);
```
and in
 ` AliDCSClient::ReceiveValueSet(TObjArray* result, Int_t& ownerIndex)`
there are the following lines:
```
  AliDCSMessage message;
  message.GetValues(result);

```
Bottom line is that AliDCSMessage::GetValues(TObjArray* result) is
called with result owning its values, against the comment in
`AliDCSMessage::GetValues`

This might be the reason why the `TObjArray` member of `AliDCSMessage`
does not own its objects.

One could then replace
`  result->AddLast(aValue);`
with
`  result->AddLast(new AliDCSValue(*aValue));`

As suggested by Ruben, for this moment it may be better to leave this code unchanged until someone can prove that it misbehaves.

PR #772, opened 2018-09-11
Return dummy track for query of ebm.bd. label in absence of subsidiary event

shahor02:
If the ESD is obtained by filtering embedded signal from the MC-2-MC production
it may contain some labels belonging to the underlaying event tracks (i.e. offsetted
by 10M). The result of the query of AliMCEvent::GetTrack(label) on such a label depends on:

1) if the galice.root of the embedding production was also distilled (the reference on the
bg. event path is removed), then a dummy AliMCParticle will be returned, made from
TParticle dummy(21,999,-1,-1,-1,-1,1,1,999,999,0,0,0,0).
Also, calls to IsPhysicalPrimary, IsSecondaryFromWeakDecay, IsSecondaryFromMaterial will return FALSE,
while IsFromBGEvent will return TRUE.
The call to GetGenerator will return TString("suppressed")

2) the galice.root refers to valid directory with the underlying event: proper MCParticle is retrieved


shahor02:
+1

PR #774, opened 2018-09-12
Place member TH1 to top root directory

shahor02:
Histograms created in some directory is destroyed when the directory is deleted. To avoid
invalid pointers the histos should be placed in the top ROOT dir.


davidrohr:
+q

davidrohr:
+1

davidrohr:
@shahor02 : Thanks for fixing this. I am wondering why we didn't see this in other components yet. probably, because there the histogram is created in the DoInit, not in the constructor. When I have time I'll scan through them.

PR #777, opened 2018-09-19
Fix for usage of on-fly V0s in AOD cascades

fprino:
These 5 commits are needed to fix an issue with the AOD cascades discussed in ALIROOT-7980.
For V0s reconstructed both on-the-fly and offline, the on-fly V0 were wrongly used in the conversion of ESD cascades (done only with offline V0s) to AOD cascades.
These commits implement:
- a fix in AliAnalysisTaskESDfilter::ConvertCascade, so that in future AODs the on-fly V0s will not be used in the conversion of cascades
- a method AliAODEvent::FixCascades that fixes current AODs by re-attach the correct (offline) V0 to the cascade. This method is executed at the first call of one of the GetCascade methods of AliAODEvent
- a flagging of fixed cascades via the fBits of AliAODcascade.


chiarazampolli:
+test

PR #780, opened 2018-09-25
Generators will tell its length/time units to AliGenMC for Boost method

shahor02:
AliGenMC::Boost() method assumes m/s for length/time units. In case the generator
uses other units it must set the AliGenMC::SetGeneratorUnitsForMeterSecond(lgtC, timeC)
with lgtC and timeC being the coefficients translating generator's units to m/s.

### DIFF
--- a/EVGEN/AliGenMC.cxx
+++ b/EVGEN/AliGenMC.cxx
@@ -33,6 +33,7 @@
 #include "AliRun.h"
 #include "AliGeometry.h"
 #include "AliDecayer.h"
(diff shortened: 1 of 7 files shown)


chiarazampolli:
Hello @rbertens , @qgp ,

Can you review this PR? We need it for the new AliRoot tag (so the sooner, the better :) ).

Thanks!

Chiara

rbertens:
+1 yes, it took me some time to look up the conversion factors ;)

PR #782, opened 2018-09-26
Revert "added missing TF1 streamers for ROOT6 compatibility"

dberzano:
This reverts commit c92257906824b486f8b1b8c32c009c6b14862719.

We suspect this commit is causing difficulties under certain conditions.
Moreover it's very wrong to have this workaround loaded as soon as `aliroot` is
called (it's not needed most of the times).


dberzano:
+1 corresponding pr is #765, we should definitely find a better way to achieve this.

PR #786, opened 2018-09-29
Remove most libdeps from the aliroot executable

dberzano:
The `aliroot` executable produced with ROOT 6 now depends solely on `STEERBase`
because of the `AliLog` class: initialization of `gAlice` is now performed
through the ROOT interpreter at runtime.

This works around a problem seeing `aliroot` stuck in an infinite mutex loop
on SLC6. The problem is likely due to a bug in an old glibc. See ALIROOT-7993.


dberzano:
+1 (@davidrohr @mazimm fyi, this works around the infinite loop problem when starting `aliroot`)

PR #797, opened 2018-10-16
Updating reader for RELDIS events

coppedis:
This commit could be included in the release to run with RELDIS code (EM dissociation events in A-A collisions), some minor changes were included to comply with the outputs provided by Igor for 5.11 TeV (the current version can be used only with inputs calculated for 2.76 TeV since some changes were introduced in between).

Thanks for the help.
Chiara


rbertens:
+1 @coppedis thanks for the update, I will give running a try in the next few days

PR #805, opened 2018-10-19
AliVertexGenFile will provide also the time

shahor02:
AliVertexGenFile used to pass the vertex from the underlying bg. event to the signal event being embedded to the bg, must pass also the fTime (smearing of the interaction time due to the finite size of the interaction diamond) used to generate the bg. event. Otherwise there is a ```sigma_VZ/c ``` time smearing between the data signal and bg. event


shahor02:
+1

PR #806, opened 2018-10-19
[EMCAL-33] Fix segfault in clusterizer component

mfasDa:


davidrohr:
+test

mfasDa:
@davidrohr In case there are more problems with the digits monitor the second commit will work around this: For some not understood reason the variable gaintype is assigned incorrectly, and I don't find an obvious memory corruption as there is no write operation in the sequence. As there are only two gaintypes I replaced this with a if-then-else clause, after which the problem disappears.

mkrzewic:
If symptoms point to memory corruption - which they do, a workaround of that type will not be enough; with enough statistics it is a guaranteed issue. Please keep looking.

davidrohr:
+1 seems to fix the segmentation faults

PR #809, opened 2018-10-22
Make AliDataFile ignore AliPhysics version tags

dberzano:
When looking for data files on CVMFS, ignore any version tag that comes after
an underscore. If version is `vAN-20181021_ROOT6-1`, then it is converted to
`vAN-20181021`: both build number and the `_ROOT6` tag are ignored.


dberzano:
+1

PR #813, opened 2018-10-24
First version of ZMQROOTprocessor

mkrzewic:
Uses a user supplied macro to process the data
will be used for VZERO calibration processing


shahor02:
+1

PR #814, opened 2018-10-24
Reduction of Output for TRD Parameters

jluehder:
Instead of giving Output for every Call of IdentifiedAsElectron, Information about missing Parameters will now only be shown once

### DIFF
--- a/STEER/STEERBase/AliPIDResponse.cxx
+++ b/STEER/STEERBase/AliPIDResponse.cxx
@@ -1776,7 +1776,7 @@ void AliPIDResponse::SetTRDPidResponseMaster()
 }
 void AliPIDResponse::CheckTRDLikelihoodParameter(){
     Int_t nTracklets=1;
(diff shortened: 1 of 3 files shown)


shahor02:
+1
Actually, it is not necessary to increase the ClassDef if there are no class data members added or changed.

chiarazampolli:
Hello @shahor02 , 

Isn't it better to revert the increase of the increase of teh ClassDef?

Chiara

PS: thanks @jluehder !

shahor02:
Well, it does not hurt, just not necessary.

Cheers
   Ruben

On Oct 24, 2018 14:10, chiarazampolli <notifications@github.com> wrote:

Hello @shahor02<https://github.com/shahor02> ,

Isn't it better to revert the increase of the increase of teh ClassDef?

Chiara

PS: thanks @jluehder<https://github.com/jluehder> !

chiarazampolli:
+1

PR #816, opened 2018-10-25
Hotfix: Check if File exists! If no TRD Response exists

jluehder:
If there is not only no Parameters in the TRDPIDResponseObject, but no TRDPIDResponseObject at all, the code crashes. 
Includes an additional if check, if fTRDPIDResponseObject exists
Also removed one AliError, for better understanding of the Code

### DIFF
--- a/STEER/STEERBase/AliPIDResponse.cxx
+++ b/STEER/STEERBase/AliPIDResponse.cxx
@@ -1780,18 +1780,20 @@ void AliPIDResponse::CheckTRDLikelihoodParameter(){
     Double_t params[4];
     Int_t centrality=0;
     Int_t iCharge=1;
-    if(!fTRDPIDResponseObject->GetThresholdParameters(nTracklets, level, params,centrality,AliTRDPIDResponse::kLQ1D,iCharge)){
-        AliInfo("No Params for TRD Likelihood Threshold Parameters Found for Charge Dependence");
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Ciao @jluehder , 

The AliInfo is printed only once, right? To me it looks fine... 

Cheers,

Chiara

jluehder:
> Ciao @jluehder ,
> 
> The AliInfo is printed only once, right? To me it looks fine...
> 
> Cheers,
> 
> Chiara

Yes it is only printed once. The problem was the following: 
SetTRDPidResponseMaster() does nothing but loading the fTRDPIDResponseObject. If the object does not exist the function gives an AliError(Form("TRD Response not found in run %d", fRun)). But this does not give the information, if there are  parameters with a charge  dependency or just combined parameter values. For example electrons and positrons would  then have the same parameters. So i wrote a function which uses this fTRDPIDResponseObject to check, if the Parameters for a certain run are included.
At this point it did not get to my mind, that the whole fTRDPIDResponseObject could possibly not exist. Of course this can happen and i therefore fixed this now.

chiarazampolli:
+1
Thanks!

jluehder:
No problem. Thanks for the approve. I just hope that now all bugs are fixed :*(

chiarazampolli:
If not, we will fix them :) 
Cheers,
Chiara

PR #818, opened 2018-10-26
Add optional parameters for external MC generators

rhaake:
Hi,

This commit allows generic optional arguments for the generator and should be fully backward-compatible. In the default setting, no functionality was changed.

The idea behind this is to give the configuration (for JEWEL) directly as parameters instead of hard-coded files. The parameters will be passed as a string to the external generator script.
I have another pull request in AliPhysics that will add the corresponding JEWEL generator files, which is tested and worked fine locally.

Best,
Ruediger

### DIFF
--- a/EVGEN/AliGenExtExec.cxx
+++ b/EVGEN/AliGenExtExec.cxx
@@ -35,6 +35,7 @@ AliGenExtExec::AliGenExtExec(const TString &scriptpath) :
   fPathFIFO("gen.hepmc"),
   fPathFile1("gen1.root"),
   fPathFile2("gen2.root"),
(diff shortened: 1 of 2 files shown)


rhaake:
Here is the connected PR in AliPhysics:
  https://github.com/alisw/AliPhysics/pull/7296

rbertens:
@qgp , what do you think? This PR is fully backward compatible, i.e. expands but does not require changes for the generator interface. Can we merge, or do we foresee a different approach?

qgp:
@rbertens: fine with me

rbertens:
+1

PR #831, opened 2018-11-04
Configurability of studied triggers

ddobrigk:


mkrzewic:
+test

mkrzewic:
is that the only thing that is needed?

ddobrigk:
@mkrzewic , it is the only urgent thing. You also need to swap the processing macro for a new version (see my mail just now). All remaining adjustments we can do only in the post-processing macro or even with the info in the logbook (and I don't want to go nuts with component changes at the last minute!) - but from what I understand configurable triggers are absolutely mandatory...

mkrzewic:
btw also here it would probably make sense to add the trigger names to the titles for reference

ddobrigk:
Hi @mkrzewic , of course a good idea. Now implemented. Thanks! 

It may also be good to have the triggers printed also in the post-processing macro titles. Will send by mail...

mkrzewic:
+1

PR #837, opened 2018-11-08
[EMCAL-409] Add MC offset for online patches

mfasDa:
For checking the online bits the MC offset needs to be taken into
account, otherwise the patches are not idenfied. Fixing filling
histograms with online patches in the PbPb trigger QA component
on the HLT.


mfasDa:
+1

Very urgent, needs still to be deployed on the HLT before the run starts

PR #841, opened 2018-11-11
fix iterator index name and lines orderning

gconesab:
Same iterator index "idig" used in loop and other loop inside used to reinitialized the trigger raw sample.
Move an MC related line before the trigger code part.


gconesab:
+test

gconesab:
+1

gconesab:
+1

gconesab:
+1

PR #844, opened 2018-11-13
Update run list in utilities.sh

mkrzewic:


dberzano:
+1 Approving, but... who is still using this? We should not use it any longer, we have decent alternatives.

PR #850, opened 2018-11-15
[EMCAL-406] Track prop to Emcal using outer param. Off by default.

loizides:
Bring the esd filter to the level that potentially one could switch to use of outer param when creating the AODs, and also implemented a switch to use mass hypothesis rather than pion mass always.
By default, all settings are like they have been used so far.

### DIFF
--- a/ANALYSIS/ESDfilter/AliAnalysisTaskESDfilter.cxx
+++ b/ANALYSIS/ESDfilter/AliAnalysisTaskESDfilter.cxx
@@ -123,6 +123,8 @@ AliAnalysisTaskESDfilter::AliAnalysisTaskESDfilter():
   fTPCaloneTrackCuts(0),
   fDoPropagateTrackToEMCal(kTRUE),
   fEMCalSurfaceDistance(440),
+  fUseMassForPropToEMCal(0),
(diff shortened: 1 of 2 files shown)


shahor02:
+test

gconesab:
wait, has alirrot been tagged with the changes in aliemcalrecoutilsbase?

chiarazampolli:
Hello,
No new AliRoot was done since v5-09-41a.
Chiara

gconesab:
@loizides, close this request until a new tag is provided.

G

shahor02:
@gconesab : not sure why do you need a new tag of aliroot to have this change, both AliEMCALRecoUtilsBase and filtering classes are in the aliroot. 
They can go to the next tag together (provided the aliphysics is not broken by these, which should not happen as far as I can see from the code).

shahor02:
@gconesab: all checks are green,  I don't see any problem with this PR.

shahor02:
+1 

To me it looks fine

gconesab:
ok, now I got it, this class  is in aliroot not aliphysics. Sorry for the noise. Night shifts are not good for mental  health.

PR #853, opened 2018-11-16
Remove TOF active mask in fired TOF trigger maps

ekryshen:
We prefer to remove TOF active mask selection both in data and MC processing to have better control over potential inconsistencies in the TOF active mask stored in DCS and actual active map used in online trigger decisions. We'll apply this mask manually in the offline analysis.

We also shift tof trigger times for LTM=64 and 65 to align them with other LTMs and avoid possible loss of hits in these channels. This approach has been proved to work in raw data analysis.


chiarazampolli:
@noferini , can you please confirm that this change is fine?

Chiara

noferini:
Dear Chiara, 
I checked with Evgeny.
The changes are fine so you can accept the pull request.
Cheers
Francesco

chiarazampolli:
+1

chiarazampolli:
Thanks Francesco!

Chiara

PR #854, opened 2018-11-19
[EMCAL-408] Add Rho value to the AOD calo trigger object

mfasDa:
Median value needs to be propagated to the AODs as well in order to be subtracted from from the raw patches at AOD level.


shahor02:
+1

dberzano:
Error unrelated, tested separately.

PR #863, opened 2018-12-03
Add New parametrization for PbPb multiplicity dependence

MartinSchmidtTuebingen:
Reworked parametrization for dEdX dependence in PbPb to parabel + constant (after minimum), follows the curve much better than before. Inserted every change needed to include this possibility in AliTPCPIDResponse. Nothing changes for the user.


shahor02:
+1

MartinSchmidtTuebingen:
The log is not found when I click on it. Also, I tested it against todays master. How can I check what is wrong?

shahor02:
Sometimes it produces a spurious error, in which case the log is not available. If by tomorrow it does not turn green, @dberzano 's intervention will be needed.

PR #865, opened 2018-12-06
Fix active time in channel masking due to HV information

wiechula:
For details see ATO-458


wiechula:
Hi @dberzano,
is the error relevant? At least it does not seem to be related to my commit.

Cheers,
Jens

chiarazampolli:
Yes i wonder the same. I saw this error in other commits too...

PR #871, opened 2018-12-12
Add HLT TRD definition, remove old friend class declaration

martenole:
- definition for TRD MC tracklet labels
- friend class declaration was for a class that is not used anymore, does not work with the new templated class


davidrohr:
+1

PR #872, opened 2018-12-14
Fix for alien file usage

miweberSMI:


rbertens:
+1 @miweberSMI thanks

PR #874, opened 2018-12-18
AliAnalysisManager : Add CreateChain static method

adriansev:
For local analysis (and not only) a static method for creating
a chain would be useful as it would bypass the need of LoadMacro
workarounds in ROOT6

### DIFF
--- a/ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
+++ b/ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
@@ -3162,7 +3162,7 @@ void AliAnalysisManager::InitInputData(AliVEvent* esdEvent, AliVfriendEvent* esd
  * @param friends Specify the root_file/friend_tree that is assumed to be in the same directory as the each input file; if friend_tree is not specified we will assume the defaults
  * @return TChain*
  */
-TChain* AliAnalysisManager::CreateChain(const char* filelist, const char* cTreeNameArg, const char* friends, Int_t iNumFiles, Int_t iStartWithFile)
+TChain* AliAnalysisManager::CreateChain(const char* filelist, const char* cTreeNameArg, Int_t iNumFiles, Int_t iStartWithFile)
 {
 TString sTreeNameArg (cTreeNameArg), treeName;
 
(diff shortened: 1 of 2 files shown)


adriansev:
@dberzano Could you please review?

ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
```diff
@@ -3139,3 +3148,92 @@ void AliAnalysisManager::InitInputData(AliVEvent* esdEvent, AliVfriendEvent* esd
...
+  if (strobj_file) { friends_filename = strobj_file->GetString(); }
+  TObjString* strobj_tree = dynamic_cast<TObjString*>(arr->At(1));
+  if (strobj_tree) { friends_treename = strobj_tree->GetString(); }
+  delete arr;
+
+  if (friends_treename.IsNull()) {
+    if (treeName.EqualTo("esdTree")) {
+      friends_treename = "esdFriendTree"; }
+    else if (treeName.EqualTo("aodTree")) {
+      friends_treename = "aodTree"; }
```
shahor02: What means aodTree as a friend for the aodTree?

ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
```diff
@@ -3139,3 +3148,92 @@ void AliAnalysisManager::InitInputData(AliVEvent* esdEvent, AliVfriendEvent* esd
...
+  TIter next(list);
+  TFileInfo* fileinfo = NULL;
+  while (( fileinfo = dynamic_cast<TFileInfo*>(next()) )) {
+    TString dirname = gSystem->DirName(fileinfo->GetFirstUrl()->GetFile());
+    TString friend_for_chain = dirname + friendinfo_for_chain;
+    if (chainFriend) { chainFriend->Add(friend_for_chain.Data()); }
+    }
+
+  if (chainFriend) { chain->AddFriend(chainFriend); }
+  }
```
shahor02: How is such chainFriend will be used? Note that the AnalysisManager reads the friends file by file (whenever the esd file in the esd chain changes) and knows how to associate the tracks in the esd and esdfriends. It also knows how to reset the friend events: the TPC seeds in the friends are using shared cluster objects which need to be properly cleared before reading next event. Reading the friends via standard trees mechanism is error prone.

shahor02:
+test

ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
adriansev: i have no idea ... the logic is from https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/macros/CreateAODChain.C#L38

ANALYSIS/ANALYSIS/AliAnalysisManager.cxx
adriansev: i do not use friends and i have no test case for them but i tried to keep the logic of https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/macros/CreateAODChain.C and https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/macros/CreateESDChain.C

adriansev: N.B. if friends option is not used this code path is not used

adriansev: Also, this is just a static function used to replace the macro that creates chain for the analysis .. so, given that AliAnalysisManager needs a chain to work on (to start analysis), i do not think that does anything automatically (i do not see any StartAnalysis signature from AliAnalysisManager that use as argument a list of files)

adriansev: well, you are the boss... i actually have this function in my task so i can use it anyway .. i just thought that would be of use to larger public and this is why i did this pr.
there are 2 options :   
i can remove the treatment of friends   
or   
just close the pr and let you do it.
imho the way i do it is more efficient and does also autodetection of the tree type (and because of TFileCollection the comments and validity of files are managed automatically)

shahor02: @adriansev : I am not the boss, I just see that the friends part of the patch may lead to crash. If you want to keep the signature you used, it is ok, but please remove the friends part.
Cheers,
 Ruben

shahor02:
+1
thanks!

adriansev:
thank you!!

PR #876, opened 2018-12-18
Feature charmpi0

mfasDa:


mfasDa:
Two changes:
- When D0 is requested to be decayed into pi0 + X the D+ is forced to be decayed into K0 pi+ p0
- The channel Lambda_c -> Lambda pi+ pi0 was missing in the decay table, therefore forcing this channel did not properly work.
As this commit affects changes in the decay table a cross check by @amorsch would be helpful.
For forcing the D+ into K0s pi+ pi0 I would like to as HF coordinators (@arossi81) for approval.
@pzhristov, @dberzano @chiarazampolli  If possible I would appreciate if the commit could still be included in the next tag in preparation for a dedicated production for HF -> Pi0 + X.

Thanks!

Cheers

Markus

chiarazampolli:
Hi Markus,

As you might understand, we won't postpone the new tag for this - but we are also not ready with the tag. So, if @arossi81 confirms, we can merge it. 
Please, open the Porting Request then.

Thanks!

Chiara

arossi81:
Hi Markus,
I am looking at it. Are you sure about the modification about line 57: is it correct?
Cheers,
Andrea

mfasDa:
Hi @arossi81,

No, this line should not have changed - thanks for spotting. I only wanted to add one decay channel. Will fix immediately.

Cheers

Markus

arossi81:
Hi Markus,
thanks! The rest is fine for me. My only observation (not much about this commit but the previous one) is that, if I well remember, the D0->K-pi+pi0 channel has a resonant sub-channel K-rho with a rather large BR. Have a look, I do not think that this is too relevant, but it may influence the decay kinematics and the kinematic of the resulting pi0.
Cheers,
Andrea

chiarazampolli:
Ok, so: should I merge, or you want to look at the comment by Andrea first?

chiarazampolli:
@mfasDa , @arossi81 , should I merge? We will tag tomorrow morning at latest (adding @fprino , @dberzano , @shahor02 to follow up, since tomorrow I will not be available)

mfasDa:
From my side ready to merge.

chiarazampolli:
+1

dberzano:
Thanks! And I am starting the validation **NOW**. Tomorrow morning is too late (if there is any issue and we need to reiterate it, everybody will be on vacation).

PR #877, opened 2018-12-20
Option to flag MC2RAW embedding

shahor02:
AliSimulation::EmbedInto(path, NSig2Bg, RAWSTATUS) will flag the embedding as being done to raw
data if RAWSTATUS==kTRUE (default is FALSE). In this case the AliMCEventHandler will not try to load the kinematics of the background event as a subsidiary MC Event


shahor02:
+1

PR #882, opened 2019-01-09
Tiny fix in median calculation

martenole:
This was causing small deviations from the O2 version of the calibration with track residuals.

### DIFF
--- a/TPC/TPCcalib/AliTPCDcalibRes.cxx
+++ b/TPC/TPCcalib/AliTPCDcalibRes.cxx
@@ -2698,7 +2698,7 @@ float AliTPCDcalibRes::MAD2Sigma(int np, float* y)
   // the input array is not modified
   if (np<2) return 0;
   int nph = np>>1;
-  if (nph&0x1) nph -= 1;
+  if (!(np&0x1)) nph -= 1;
   // don't abuse stack
   float *ycHeap=0, ycStack[np<kMaxOnStack ? np:1],*yc=np<kMaxOnStack ? &ycStack[0] : (ycHeap = new float[np]);
   memcpy(yc,y,np*sizeof(float));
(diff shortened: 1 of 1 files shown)


miranov25:
Could Ruben approve, or confirm change? Checking the code change how much the RMS estimate changed?

TPC/TPCcalib/AliTPCDcalibRes.cxx
```diff
@@ -2698,7 +2698,7 @@ float AliTPCDcalibRes::MAD2Sigma(int np, float* y)
   // the input array is not modified
   if (np<2) return 0;
   int nph = np>>1;
-  if (nph&0x1) nph -= 1;
+  if (!(np&0x1)) nph -= 1;
```
shahor02: @martenole: thanks for spotting, but actually, the bug is in the presence of this line, it is simple not needed. I.e. the ``nph=k`` both for the odd ``np = 2*k+1`` and even ``np = 2*k``. 
The median is computed as the element y[k] (of sorted array) for odd np (i.e. central element is taken, e.g. nph=1 for np=3) and the mean of 2 central elements y[k-1] and y[k] for even np (e.g 0.5*(y[1]+y[2]) for np=4). Please, just eliminate this line.

BTW, I've just noticed, that the ``0.5f*(SelKthMin(nph-1,np,yc)+SelKthMin(nph,np,yc))`` can be calculated faster as ``0.5f*(SelKthMin(nph-1,np,yc)+SelKthMin(0,nph,yc+nph))``, since the ``SelKthMin(k...)`` places all elements ``< k-th`` before it and those ``> k-th`` after it. For large np (num.recipe suggests ~100) the difference between the odd and even cases becomes irrelevant, and we can use faster odd case recipe. For the AliRoot this is already irrelevant, but for the O2, could you use for the median
```
int nph = np>>1;
median = (np&0x1 || np>100) ? SelKthMin(nph,np,yc) : 0.5f*(SelKthMin(nph-1,np,yc)+SelKthMin(0,nph,yc+nph));
```

shahor02:
@martenole thanks, please see the comment I've added to the patch.

PR #883, opened 2019-01-09
fix in TOF SDigitizer for rare problematic hits

noferini:
Fix to correct rare hits with a residual ouside the pad (keep them inside)


noferini:
We noted a problem during a recent simulation with geant4. In some rare cases the hit coordinate inside a sensitive volume were slightly outside (1.25001 > 1.25). This produced a wrong mapping of the volume during sdigitization and then a Fatal Error in the digitization.
We added a check to avoid coordinates larger than the pad size.

chiarazampolli:
+1

PR #884, opened 2019-01-10
Fix index for median calculation

martenole:
Sorry for another PR and the spam, I did not know how to change the commit from the previous PR. As suggested by Ruben this line simply needs to be deleted.


shahor02:
+1

PR #885, opened 2019-01-10
Fix: Not incrementing event number (fNcurrent) in NextParticle

miweberSMI:


rbertens:
+1 @miweberSMI should it be checked if this increment is also present in the single EMD class?

miweberSMI:
Hi @coppedis, Igor Pshenichnov,

I am not sure if I am right, but I think the incrementing of event number (fNcurrent) needs to be removed in the NextParticle call? If I do this, my tests (starting at event number n and do one event and still getting same results as in the last event when starting at 0 and doing n events) are OK. W/o this fix I got not compatible results. Please let me know, what you think, so that @qgp and @rbertens can approve this PR.

Thanks and cheers
Michael

miweberSMI:
Hi @rbertens,

this is actually the single EMD class ;). But I checked also the other one and I didn't find it, but perhaps good if somebody could confirm. I did not test it, because we only have a file in the PWG-MM folder for the single events...

Thanks and cheers,
Michael

coppedis:
Dear @miweberSMI,
sorry I did overlook something: you are certainly right: fNcurrent is in fact incremented 2 times in NextParticle() (wrongly!) and in NextEvent()...therefore your fix is needed.
Thanks a lot, 
cheers ch.

PR #886, opened 2019-01-14
Rename TRD tracking components AliHLTTRD* -> AliGPUTRD*

martenole:
This commit requires also a patch in AliTPCCommon (https://github.com/AliceO2Group/AliTPCCommon/pull/16)


davidrohr:
Needs to be merged together with an alidist PR, and with the creation of a new AliRoot tag. I'll create a JIRA ticket.

davidrohr:
We can close this and use alisw/AliRoot#888 instead, which contains the changes for both TPC and TRD.

PR #900, opened 2019-02-15
Add prototype of ZMQESDserver

mkrzewic:
Adds tool (standalone binary, ZMQESDserver) to read ESDs and send the extracted columnar data via ZMQ in O2 data format. Meant to inject AOD like data into O2.
Will probably need some further development regarding the precise format of the data. Otherwise works fine.

### DIFF
--- a/HLT/BASE/util/AliHLTZMQhelpers.h
+++ b/HLT/BASE/util/AliHLTZMQhelpers.h
@@ -111,8 +111,8 @@ struct AliHLTDataTopic : public DataTopic
 };
 
 class AtomicFile {
-  TString targetFileName;
-  TFile* tempFile;
+  TString targetFileName{};
+  TFile* tempFile{nullptr};
   public:
   AtomicFile(const char* name);
   ~AtomicFile();
--- a/HLT/BASE/util/CMakeLists.txt
+++ b/HLT/BASE/util/CMakeLists.txt
@@ -132,6 +132,8 @@ if(ZEROMQ_FOUND)
                       ${ZEROMQ_INCLUDE_DIR}
                       )
 
+  add_executable(ZMQESDserver ZMQESDserver.cxx)
+  target_link_libraries(ZMQESDserver HLTbase AliHLTGlobal AliZMQhelpers RAWDatabase Geom Graf MathCore Net Tree EG Gpad Matrix Minuit Physics VMC Thread STEERBase XMLParser Graf3d RIO Hist Core pthread)
   add_executable(ZMQfileProxy ZMQfileProxy.cxx)
   target_link_libraries(ZMQfileProxy HLTbase AliHLTGlobal AliZMQhelpers RAWDatabase Geom Graf MathCore Net Tree EG Gpad Matrix Minuit Physics VMC Thread STEERBase XMLParser Graf3d RIO Hist Core)
   add_executable(ZMQROOTmerger ZMQROOTmerger.cxx)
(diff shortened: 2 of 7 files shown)


chiarazampolli:
ciao @mkrzewic , 

I am not the best to approve this commit... @dberzano , @shahor02, should I?

Cheers,

Chiara

mkrzewic:
@chiarazampolli it should not interfere with anything since it lives in the HLT zeromq subsystem - probably will not even get compiled by default (unless libzmq is detected).

sawenzel:
+test

sawenzel:
This should be ok. Needed as input for Run2. @mkrzewic : Can you add a simple docu/tutorial how to run this and feed data into DPL workflow?

sawenzel:
And I would be in favour of compiling the code (and potentially adding zeromq as a dependency).

mkrzewic:
@sawenzel i'll add a short howto. Integration with DPL is something i'll start looking into right now - in principle it should work out of the box minus the Arrow integration - for that i'll see how to slap arrow metadata on a bunch of raw arrays.
As for compilation: it compiles once it detects zeromq - adding that as dependency takes care of that.

ktf:
Besides the missing DPL support as pointed out by @sawenzel, I am still not convinced about this approach:

* It seems to me that in order to avoid a plain copy of a 300 files and adding 50 lines of code for the converter in O2, we are duplicating the functionality which we currently have in O2, reimplementing 500+ lines of new code.
* It relies on ZeroMQ directly, rather than going through FairMQ. While it's true that the two are compatible when using ZeroMQ, are we really sure that's the backend we will need to use? Put more directly, how will this play if we want to use shared memory or any other kind of transport we might be required to use (e.g. when testing on the vertical slice)?
* Anyone who wants to use this on his own machine will still have to compile the full AliRoot. 

Concerning this last point, why don't we extract the set of classes needed in a standalone package? We could use my script to keep it automatically synched with AliRoot on every commit and it would still be completely decoupled from O2 so that if people do not want to use it, don't have to.

mkrzewic:
about "DPL support" - this device provides plain data, it is not up to the data to support some framework, the framework should support plain data (and it does, to a degree, AFAIK).
The 600 lines of code looks more work than it actually was - it was just a copy/paste from existing HLT devices, I just never felt the need to formalize the boiler plate into a "framework" for the limited number of cases we actually had a separate device "classes" running zeromq:)
Anyway - being able to work with plain data in O2 is important in general and I will keep developing this interface for the time being.

mkrzewic:
@sawenzel I added an md file (ZMQESDserver.md) in the same dir as a start. I consider this initial version of the aliroot side done for now. I'll see what can be done on the O2 side.

sawenzel:
This was discussed as an option worth investigating in the last O2 WP3 meeting. Merging.

sawenzel:
+1

ktf:
Ok, but then I think it would be fair to merge also my other PR, with optional compilation, to compare the plus and cons. I will update it and merge it in O2.

ktf:
Also, I guess you mean O2 WP4.

sawenzel:
Sure, I meant WP4.

sawenzel:
The log cannot be inspected. 404.

aphecetche:
@mkrzewic is there a JIRA ticket about this one where we can discuss it further ? (e.g. I see it's now only exporting a limited part of the ESD, I guess this is just because it's a WIP ?)
@kft I like the idea of having a separate package. Which would satisfy the requirement of not having to drag along full AliRoot and is in addition not incompatible with the ZMQESDServer approach, btw.

mkrzewic:
@aphecetche sure, the idea is to add stuff as we need it.
About the JIRA ticket: we can create one, question is where? We have a multitude of jira projects...

ktf:
https://alice.its.cern.ch/jira/projects/O2/issues, I think @Barthelemy is trying to consolidate everything there.

Barthelemy:
Hi, yes, the idea is that by default things go to O2 Project. People can have their more specific JIRA projects but if you don't know, or don't care, just put it in O2 project. We can add more components in a blink if you need.

mkrzewic:
ok, for reference the JIRA ticket is https://alice.its.cern.ch/jira/browse/O2-510

PR #903, opened 2019-02-18
remove cause for format-truncation warning

zwound40:
on gcc version >=7, the snprintf call caused a format-truncation warning. By default, -Werror is used, so the warning prevented compilation.

### DIFF
--- a/STEER/STEER/AliLHCDipValT.h
+++ b/STEER/STEER/AliLHCDipValT.h
@@ -117,8 +117,8 @@ Char_t* AliLHCDipValT<Element>::TimeAsString(double t, Bool_t utc)
   static char buff[22];
   time_t tt = (time_t) t;
   struct tm *time = utc ? gmtime(&tt) : localtime(&tt);
-  if (snprintf(buff,21,"%02d:%02d:%02d %02d/%02d/%04d",time->tm_hour,time->tm_min,time->tm_sec,
-	  time->tm_mday,time->tm_mon+1,time->tm_year+1900) ) return (char*)buff;
+  if ( snprintf(buff,21,"%02d:%02d:%02d %02d/%02d/%04d",time->tm_hour,time->tm_min,time->tm_sec,
+	  time->tm_mday,time->tm_mon+1,time->tm_year+1900) > 0 ) return (char*)buff;
   return 0x0;
 }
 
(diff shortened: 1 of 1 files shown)


STEER/STEER/AliLHCDipValT.h
```diff
@@ -117,9 +117,9 @@ Char_t* AliLHCDipValT<Element>::TimeAsString(double t, Bool_t utc)
   static char buff[22];
   time_t tt = (time_t) t;
   struct tm *time = utc ? gmtime(&tt) : localtime(&tt);
-  snprintf(buff,21,"%02d:%02d:%02d %02d/%02d/%04d",time->tm_hour,time->tm_min,time->tm_sec,
-	  time->tm_mday,time->tm_mon+1,time->tm_year+1900);
-  return (char*)buff;
+  if (snprintf(buff,21,"%02d:%02d:%02d %02d/%02d/%04d",time->tm_hour,time->tm_min,time->tm_sec,
+	  time->tm_mday,time->tm_mon+1,time->tm_year+1900) ) return (char*)buff;
```
shahor02: Hi @zwound40 , under which circumstances this snprinf can return 0? As far as I see it can never happen

zwound40: Hi @shahor02 , sorry this was a failure on my side. sprintf will return a negative number in case of failure, not 0. I committed a fix.

shahor02: @zwound40 OK, thanks, this makes more sense. But did you encounter a situation when the snprintf has failed?

shahor02:
+1

STEER/STEER/AliLHCDipValT.h
zwound40: @shahor02 : No, the only reason for this fix it that gcc (tested with gcc 8.0.2) gives a format-truncation warning, if the return value is not used. Since the code is by default compiled with -Werror this results in the compilation failing.  One could maybe even do something like:
int dummy = sprintf(...);
return return (char*)buff;
with a corresponding comment in the code that this is simply a hack to make gcc happy.

shahor02: OK, thanks.

PR #907, opened 2019-02-25
Fix behaviour when 1st generated particle has unknown PDG (R. Shahoyan)

chiarazampolli:
This is needed for HIJING, e.g. when particle 3124 is the first (and
its PDG is unknown), then the track is not added, and we should
account for this.


chiarazampolli:
+1

PR #911, opened 2019-02-28
Access to reaction plane angle

amorsch:
A user (Takumi Osako ‎[osako@quark.hiroshima-u.ac.jp]‎ ) has asked for access to the reaction plane 
angle in HIJING.


chiarazampolli:
+1

PR #916, opened 2019-03-08
 decaytable: first try to use local version 

miweberSMI:


miweberSMI:
Dear all, 

I am pinging also @amorsch on this PR, so that he could comment.

Thanks and cheers,
Michael

fprino:
Hi,

I was thinking that we could try to modify openDecayTable.F and AliDecayerPythia in such a way that we can have the name of the file with the decay table as a data member of AliDEcayerPythia, which can be set from the Config macro, and pass it to openDecayTable.F the name .
In this way, we could have different decay table files committed in AliRoot/data (e.g. decaydatable-lowMassee.dat) and use them in the MC productions by changing the Config file.

What do you think?

Cheers, Francesco

miweberSMI:
Hi @fprino,

that would be of course much more elegant. I did the poor man's solution here...

Cheers,
Michael

amorsch:
+1

chiarazampolli:
+1

PR #918, opened 2019-03-11
Add case to decay both beauty and charm hadrons

ffionda:


pzhristov:
+1

dberzano:
@Atlantic777 is having a look, macOS failure relates to alisw/alidist#1503

PR #921, opened 2019-03-19
MTR: fix unitialised values reported by valgrind

bovulpes:
valgrind reports "Conditional jump or move depends on uninitialised value(s)", here are two fixes.
 
One of them could come from a bug:

 AliMUONTriggerGeometryBuilder::BuildChamberTypeE

Double_t dpar[3] = {0., 0., 0.};

why not initialised?

### DIFF
--- a/MUON/MUONgeometry/AliMUONVGeometryBuilder.cxx
+++ b/MUON/MUONgeometry/AliMUONVGeometryBuilder.cxx
@@ -23,6 +23,8 @@
 // 23/01/2004
 //-----------------------------------------------------------------------------
 
+#include <cassert>
+
 #include "AliMUONVGeometryBuilder.h"
 #include "AliMUONGeometryModule.h"
 #include "AliMUONGeometryDetElement.h"
@@ -288,7 +290,8 @@ void AliMUONVGeometryBuilder::SetTransformation(Int_t moduleId,
     = TGeoCombiTrans(translation, rotation);
 
   // Apply frame transform
-  TGeoHMatrix newTransform = ConvertTransform(transformation);
+  TGeoHMatrix newTransform;
+  assert(newTransform = ConvertTransform(transformation));
 
   // Set new transformation
   geometry->SetTransformation(newTransform);
--- a/MUON/MUONsim/AliMUONTriggerGeometryBuilder.cxx
+++ b/MUON/MUONsim/AliMUONTriggerGeometryBuilder.cxx
@@ -1623,7 +1623,7 @@ void AliMUONTriggerGeometryBuilder::BuildChamberTypeE(Int_t& iVolNum, Int_t icou
     Float_t zRatio = AliMUONConstants::DefaultRatioTriggerChamber(icount);      
     Float_t xEnv = (fgkDXZERO+fgkXMAX/2.)*zRatio;
    
-   Double_t dpar[3];    
+   Double_t dpar[3];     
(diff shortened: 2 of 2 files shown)


MUON/MUONgeometry/AliMUONVGeometryBuilder.cxx
```diff
@@ -288,7 +290,8 @@ void AliMUONVGeometryBuilder::SetTransformation(Int_t moduleId,
     = TGeoCombiTrans(translation, rotation);
 
   // Apply frame transform
-  TGeoHMatrix newTransform = ConvertTransform(transformation);
+  TGeoHMatrix newTransform;
+  assert(newTransform = ConvertTransform(transformation));
```
aphecetche: I guess the assert inclusion is a left over from some tests ?

shahor02:
+test
@bovulpes I wrote just yesterday to @aphecetche  about exactly this bug, but I was not sure if the initialization of dpar to 0 is proper thing to do. Leaving to Laurent to comment.

aphecetche:
@shahor02 yes, your report triggered some checks by Bogdan and Diego. The thing is it's still unclear whether a simple {0,0,0} initialisation is the right thing to do (leaving to Diego to comment further)

MUON/MUONsim/AliMUONTriggerGeometryBuilder.cxx
```diff
@@ -1623,7 +1623,7 @@ void AliMUONTriggerGeometryBuilder::BuildChamberTypeE(Int_t& iVolNum, Int_t icou
     Float_t zRatio = AliMUONConstants::DefaultRatioTriggerChamber(icount);      
     Float_t xEnv = (fgkDXZERO+fgkXMAX/2.)*zRatio;
    
-   Double_t dpar[3];    
+   Double_t dpar[3] = {0., 0., 0.};
```
dstocco: Looking at the code, it seems that the RPC type D, E, and F are the RPCs of type "long". ~
For some reason they where built separately, but one could have simply built the base shape and then properly place them in space. Indeed, AFAIK, the type D and F only differ by the y position, while the type E differ from the other also for Z position.
Anyways, in type D and F the dpar is initialised to the values:
dpar[0] = (fgkXMAX/2.)*zRatio;
dpar[1] =  fgkYMIN*zRatio;
dpar[2] = 0.4;
I have to say that I do not fully understand the meaning of those values, but I guess that the type E should have the same initialisation.

bovulpes:
At the end, from the point of view of the values of those parameters, chambers D, E and F are the same, A is different and B is different.

MUON/MUONgeometry/AliMUONVGeometryBuilder.cxx
bovulpes: Similar warnings of uninitialised variables, when TGeoMatrix constructors are called, come from other parts of the MUON geometry builder, not only the trigger part. A common strategy should better address those issues, this is kind of rapid fix to understand what the problem is.

aphecetche: @bovulpes I have two issues with this line. 

- I don't understand what it is supposed to test ? 
- also, that's dangerous to execute code in an assert, as the assert is disabled in release mode (i.e. when NDEBUG is defined)

bovulpes:
Closing for further investigation.

PR #923, opened 2019-03-22
With embedding MCEvent::GenEventHeader combines all headers

shahor02:
In case of embedding AliMCEvent::GenEventHeader() will return AliGenCocktailEventHeader 
(downcasted to AliGenEventHeader) containing the headers of both embedded and background events. 
In order to correspond to primary tracks ordering and make work methods like GetGenerator(itrack), 
the headers of the embedded event are added first, followed by the BKG event headers (e.g. Hijing).


shahor02:
+1

PR #924, opened 2019-03-25
AliDirList

mgheata:
This PR introduces a special container AliDirList that allows writing its content into a folder and reading content from a folder in a lazy manner. This container is useful for handling large outputs of analysis tasks during the grid merging phase and after, when running the Terminate method. To use, simply replace the output type of a task from TList/TObjArray to AliDirList.

The output content of a task having a AliDirList output will be placed in a folder named the same as the data container. Note that the AddTask macro for such a task needs to be changed to create a container of type AliDirList instread of the usual TList/TObjArray. During merging, the memory footprint will be reduced to the size of the largest object being merged, instread of the size of the full list. During the Terminate method, the AliDirList will be created from the merged file by scanning the appropriate folder. Note that the list will appear as having all objects inside, but these will be read from file only when actually accessed (lazy reading). This will also minimize the memory use during the Terminate phase.

The PR is fully backward compatible, not affecting any tasks not using this special container.

### DIFF
--- a/ANALYSIS/ANALYSIS/ANALYSISLinkDef.h
+++ b/ANALYSIS/ANALYSIS/ANALYSISLinkDef.h
@@ -15,5 +15,6 @@
 #pragma link C++ class  AliAnalysisTaskCfg+;
 #pragma link C++ class  AliAnalysisFileDescriptor+;
 #pragma link C++ class  AliXMLParser+;
+#pragma link C++ class  AliDirList;
 
 #endif
--- a/ANALYSIS/ANALYSIS/AliAnalysisDataContainer.cxx
+++ b/ANALYSIS/ANALYSIS/AliAnalysisDataContainer.cxx
@@ -95,6 +95,7 @@ AliAnalysisDataContainer::AliAnalysisDataContainer(const char *name, TClass *typ
 /// Default constructor.
 
    SetTitle(fType->GetName());
+   if (fType->InheritsFrom("AliDirList")) SetDirList(true);
 }
 
 //______________________________________________________________________________
@@ -462,7 +463,9 @@ void AliAnalysisDataContainer::ImportData(AliAnalysisDataWrapper *pack)
       fDataReady = kTRUE;
       // Imported wrappers do not own data anymore (AG 13-11-07)
       pack->SetDeleteData(kFALSE);
-   }   
+      // Imported data is read from file, the container must own it
+      SetDataOwned(true);
+   }
 }      
       
 ClassImp (AliAnalysisDataWrapper)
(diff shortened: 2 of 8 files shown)


dberzano:
+test

chiarazampolli:
Ciao Mihaela, 

Thanks for the implementation, it looks very useful indeed!
Will this be advertised to the analyzers once merged? I guess so, but I wanted to be sure. 

Cheers,

Chiara

miranov25:
Hello @mgheata , @chiarazampolli and Mesut ()

I'm not sure what is the main purpose of this modification, but what we considered to be important is merging per "selected" components.
We were discussing with Mesut about enabling alihadd functionality in lego trains, to make merging objects in user defined groups. 
In my opinion, this will be solution for merging problem.

Marian

chiarazampolli:
Ciao Marian, 

I don't know the context of this development, I saw it, and since it can be useful, I think it should be advertised. But I believe the use case was different from what you mention here. Mihaela can comment better.

Cheers,
Chiara

mgheata:
Hi,
This utility is not a final solution to the merging problem, but will reduce the memory footprint in case of tasks using large lists of histograms (not TTree objects). The objects are written in folders rather than in single key lists, so during merging, the peak memory will scale with the size of the largest histogram in the list rather than the size of the full list. This is particularly useful if the output contains THNSparse or large TH2/TH3. Also, the memory in Terminate is much lower since only the histograms actually used will be read from the merged file. This is just an optional feature for analyses with heavy histograms lists, not affecting the existing framework or tasks. Of course it will be advertised after some tests in grid.

miranov25:
Hello @mgheata  and @chiarazampolli 

I think I understand what you did and I agree this work is important.

What I was referring was "merging" per groups,similar  as we are doing using 
**alihadd tool**  and extension **alihadd.sh**
* alihadd 
````
alihadd
usage: alihadd [-h] [-f[kf123456]] [-a] [-k] [-T] [-O] [-n <max numer of files>] [-v [<level>]] <output file> [@]<input file> [<input file>]
````
  * alihad important option:
````
  -i <regex>
      Include TKeys whose name matches <regex>. This option can be used multiple times.
   -s <max size of file>
      Specify the approximate size of an output file.
````
* alihadd.sh extension (combining functionality ifalihadd+alisync)  
  * additional option: **checkKeys, prefetch, timeOut**
  * Example use - merging non corrupted highPt trees
    * this code need better documentation - we pleaned to write it with Mesut
````
alihaddsh -k -s 1000000000   -checkKeys highPt -prefetch 1 -v 1 -timeOut 300 test.root -inputList input.list
````

Having such option in the lego train, enabling  simplifying merging per group solve the problem with merging of lego trains.  Currently we are doing the merging of the trees, by hand ... using alihadd
We were discussing with Mesut to enable the tool also for lego trains.

Regards
Marian

jgrosseo:
This looks very useful.

AliDirList seems to act as a collection. Shouldn't it be derived from TCollection?

mgheata:
I looked at TCollection and it seems a too heavy class. In principle it would be possible, but I don't see the gain in doing that.

pzhristov:
+1

PR #927, opened 2019-03-27
Fix: allow compiling AliRoot without AliEn

Atlantic777:
This used to be possible but was broken when AliEn-ROOT-Legacy was extracted from ROOT. When AliEn is missing, just print a warning instead of throwin a fatal error.


davidrohr:
+1 Tested locally and works as expected

dberzano:
+test

pzhristov:
+1

PR #620, opened 2018-03-16
Fast dipole parameterization data

syamasak:
related: https://github.com/alisw/AliRoot/pull/552
This PR contains dipole B-field parameterization data (C++ sources and headers)

PR #928, opened 2019-03-31
Merge fast dipole data (Again, improved)

syamasak:
This PR tries to merge the AliMagFast dipole data into the upstream.

The data was not merged at the time because there was a revert due to some mysterious behavior of a build bot (#620): the previous PR was merged even though the test was failed... perhaps a build process was aborted by OOM killer since the input is very large (~40k LOC.)

An improvement of this PR is partitioned source files that should allows compiler to build with lower amount of memory. So it would be less likely to be killed or be panic.

Successfully built with alidock on my machine.
Could you check this branch works correctly? Thanks.


syamasak:
Hi Ruben (@shahor02), thanks about the long wait. This is the promised commit.
Could you check it when you are avaliable?

syamasak:
FYI: Partitioning is done with the following bash.
https://gist.github.com/syamasak/beb3a83b56b6f22ab9912c96f867e295

shahor02:
+test

shahor02:
Hi @syamasak 

Thanks a lot for your time! Will check as soon a possible.

Cheers,
 Ruben

shahor02:
Hi @syamasak ,
the files are still big... Do you understand why there is such a big difference in size of 2K and 5K fields?
The real fields should not be significantly different, just by the contribution from the solenoid return flux.

@dberzano : these are not binary data files but automatically generated code, for CPU speed reason they contain a huge set of polynomial functions + special structures for the parametrization patch search. The [recipe](http://alisw.github.io/git-advanced/#how-to-use-large-data-files-for-analysis) will not work here...

dberzano:
@shahor02 if they are generated, why don't we generate them during the build process then? We don't need to store them here.

dberzano:
For the record I think that having the method to generate those long snippets built in our code is even better. What I am asking is whether there is a hard constraint to _not_ have it here, such as it takes way too much time to generate the sources (how much?) or so.

shahor02:
Hi @dberzano, they are generated from large [root file](https://github.com/alisw/AliRoot/blob/master/data/maps/mfchebKGI_meas.root) already stored in the aliroot (and O2) using dedicated haskell code. So, if we go for the on-the-fly generation, we will need at least to introduce the haskell as a dependency. As for the generation speed, @syamasak may comment, I suspect it takes some time.

Cheers,
 Ruben

dberzano:
I see. It's maybe better not to introduce Haskell as a dependency here. I am still curious (this is, I repeat, just out of curiosity!) to know from @syamasak about the generation time.

For the record: the protection kicks in in case the PR diff size exceeds 20 MB, which is quite a lot already, but OK, I'll force-merge it, I understand the use case.

@shahor02 Shall I proceed?

syamasak:
Hi @shahor02, 
> Do you understand why there is such a big difference in size of 2K and 5K fields?
This is (maybe) because 2k params have smaller chebs against 5k, and my algorithm try to (naively) search an equally spaced grid which covers every border of the chebs to build the search index. This method may be improved by adding some translation to the grids.

Hi @dberzano, haskell code is introduced just for fast prototyping and what going on is not so hard. So it is possible to replace it with pure C/C++.
However, one more dedicated code is used to generate `*.h`. This code runs on Mathematica and depends on its `HornorForm` and `Simplify` function to performe offline formula optimization.
Of cource, the execution requires one valid liscense...😢

Code generation is slow since it is oneshot task and emphasis was placed on fidelity and runtime speed. First stage (haskell one) takes tens of minutes for 2k and 10+ hours for 5k (the cost is exponential of a number of grid division).

syamasak:
> It's maybe better not to introduce Haskell as a dependency here.

I think so too. The dependency is large and download is slow. It should be avoided.

dberzano:
OK thanks for clarifying, and thanks for your PR!

syamasak:
Thanks for merge!

shahor02:
Hi,

Just saw that it is merged... In principle, I wanted to test it 1st, but ok, in any case it will not affect aliroot operation. Thanks!

Cheers,
 Ruben

PR #931, opened 2019-04-03
Fixing problem in AOD filtering for ZDC

coppedis:
Code fix to prevent some rare cases in which the AOD filtering for the ZDC was writing wrong timing in AODs. Code checked on data with Francesco Prino.
Thanks, 
chiara


chiarazampolli:
+1

chiarazampolli:
Ciao,
There are many changes in this commit, but mainly blank spaces. The important piece is the one we discussed via mail about the initialization of ZDC variables during the ZDC AOD conversion. I think nothing else was touched (even by mistake). So I approved. 

Chiara

PR #936, opened 2019-04-08
Remove broken root-config query

mpuccio:
With the new alien-root-legacy one has to check if the TAlien class
is available instead of querying root-config (anyway it was bad design).


mpuccio:
Btw, I cross checked with the O2 root installation (that does not have alien enabled) and this check is effective:
```cpp
root [0] TClass::GetClass("TAlien")
(TClass *) nullptr
```

dberzano:
+1 thanks, the original check was very weak!

PR #937, opened 2019-04-08
Increased line size when parsing MLTrainConfig files.

mgheata:
This fix allows reading longer lines from the train configuration files in the LEGO test (10000 characters maximum instead of 1024)


pzhristov:
+1

PR #938, opened 2019-04-09
AliMC: volume-to-module map; deterministic event seeding

sawenzel:
* Simulation now writes a text file containing a mapping
  of logical volume names to the module that owns them;
  This is needed for accounting of MC steps to modules (as done
  in the MCStepLogger and related analysis)

* Adding an option to achieve reproducible/deterministic event sequence
  when setting a special environment variable. This is a useful and painless way
  to achieve same event sequence for G3 and G4 productions, for instance.

### DIFF
--- a/STEER/STEER/AliMC.cxx
+++ b/STEER/STEER/AliMC.cxx
@@ -22,7 +22,7 @@
 //         Federico.Carminati@cern.ch
 
 #include <string.h>
-
+#include <fstream>
 #include <RVersion.h>
 #include <TArrayI.h>
 #include <TClonesArray.h>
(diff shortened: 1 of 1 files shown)


sawenzel:
+test

sawenzel:
+test

sawenzel:
+test

shahor02:
+1

dberzano:
@sawenzel the tests do run the gun tests, both on macOS and Linux. I have checked.

sawenzel:
@dberzano : This is not evident to me. I can't see this in the logs. Would the system report it if the gun test failed?

dberzano:
If the guntest fails, the logfile will be displayed and a link to the logfile will be present, as for any other package, so yes.

sawenzel:
Ok. Thanks for the info.

PR #943, opened 2019-04-10
AliESDtrackCuts: added pt dependent cov matrix cut

mlknichel:
Added the possibility to cut on the covariance matrix relative pt resolution in a pt dependent way


jgrosseo:
+1

mlknichel:
seems that the automatic check tries to test the two commits in separately, like this will never work because both single commits do not compile, only if both are applied at the same time...

PR #954, opened 2019-05-19
protection against 0 division in old MC prods

gconesab:


chiarazampolli:
+1

chiarazampolli:
Ciao @gconesab ,

Do we need to keep track in JIRA that this change will go to the next tag?

Cheers,

Chiara

PR #955, opened 2019-05-21
print error message only once

jgrosseo:


jgrosseo:
@pzhristov Could you have a second look even if I only changed printouts? Thanks!

PR #958, opened 2019-05-28
Fix missing adaption to merging of AliGPU

davidrohr:
Was simply forgotten to take into account when we merged AliTPCCommon, and the old code didn't break anything but just disabled some HLT TRD components.


davidrohr:
+1

PR #960, opened 2019-06-04
fix treatment of B=0 track extrapolation X estimates

shahor02:
relevant only in some marginal cases of B=0 data, does not affect productions


shahor02:
+1

PR #965, opened 2019-06-25
Added bach-baryon cospa to AliAODcascade for nanoAODs

ddobrigk:
This commit is necessary to make full use of AODs for the cascade analysis in Pb-Pb, which is badly needed to avoid excessive CPU usage by LEGO trains. The change is essentially the addition of an extra data member to AliAODcascade that stores the bachelor+baryon CosPA such that the analyser can select / reject based on that variable at the AOD level. To be tested. Please check this code for correctness, indeed. Thank you!

### DIFF
--- a/STEER/AOD/AliAODcascade.cxx
+++ b/STEER/AOD/AliAODcascade.cxx
@@ -35,7 +35,8 @@ AliAODcascade::AliAODcascade() :
   
   fMomBachX(999),
   fMomBachY(999),
-  fMomBachZ(999)
+  fMomBachZ(999),
+  fBachBaryonCosPA(0)
    
 {
   //--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jgrosseo:
@ddobrigk The changes are fine with me. Does this compile against the current AliPhysics? (changed constructors)

ddobrigk:
@jgrosseo yes, I tested it against my local AliPhysics installation and everything works fine. I took care to change the default constructors such that the new variable is only at the end and has a default value so that any old call to the constructors will work - but it certainly won't hurt to +test this. I also didn't forget to change the classdef (would be scary if I didn't, I guess. The changes are simple but the risk is great...) 

Note also that for this to be fully functional a small change to the ESDfilter class is needed too. I'll look into that ASAP (sorry for the delay) so that we can test. Thanks!

jgrosseo:
+1

jgrosseo:
@ddobrigk  The test will run anyway, so I +1'd it, but it will not test against AliPhysics (just AliRoot). I trust your test then. :)

PR #970, opened 2019-07-02
Fix issue for multiplicity correction

MartinSchmidtTuebingen:
negative or zero values for the curvature of the multiplicity correction function should not happen. Because they occur anyway, it causes a break of the correction factor. This introduces a fix


chiarazampolli:
+1

PR #974, opened 2019-07-08
Corrects TRU ADC mapping, Fixing Missing Patch issue in Simulation

moliver813:
Corrects the ADC mapping for the TRU objects created here.
This fixes the mapping error observed through trigger patches.
See Ticket [EMCAL-509 ](https://alice.its.cern.ch/jira/browse/EMCAL-509)


mfasDa:
Please squash the commits into 1 single commit!

mfasDa:
+1

PR #975, opened 2019-07-08
Improvements to ITS geometry to check material budget discrepancies

mario6829:
Some details in the ITS geometry descriptions were improved to check the discrepancies in material budget between data and MonteCarlo. Namely:
- the copper wire running above all SDD ladders was removed
- the material inside SPD cooling pipes was changed from Liquid Freon to Gaseous Freon
- the so-called Omega Shape inside the SPD Thermal Shield was reimplemented (it was part of a complex CompositeShape, now it is a standalone volume)


mario6829:
@ktf I just created the Pull Request, let me know in case of problems. Cheers
Mario

ktf:
+test

PR #976, opened 2019-07-08
Corrects the setting of the patch size for L1 Jet Trigger.

moliver813:
Now set based on OCDB patch value, not firmware version.
Old behavior for Run 1.


mfasDa:
+1

As discussed this commit is OK as it uses the validated settings for run1 and the updated ones for run2 and should be merged.

PR #978, opened 2019-07-11
Updated deacyer configs for Eta Prime MC generator

fkellere:
Changed eta prime decayer config to include also a pi+pi-eta decay, in addition to the gamma gamma decay, with a 50:50 branching ratio.


shahor02:
+1

PR #981, opened 2019-07-16
Fix to allow to trigger on antiparticles also for HF signals

fprino:
This modification allows to trigger on particles with negative PDG code also for HF.
With the previous code, the trigger particle PDG code was checked with the absolute value also in the HF related event selections.
As a consequence, in HF productions only particles were selected as trigger particles.
With this fix one can select particles or antiparticles as triggers.


chiarazampolli:
+1

dberzano:
@chiarazampolli man you're too fast 😄

chiarazampolli:
I was checking my mails in that exact moment :-)

PR #983, opened 2019-07-21
Change ESD->AOD filter to add BachBaryon CosPA automatically

ddobrigk:
This change is meant to add one extra Float_t to each AliAODcascade during ESD->AOD filtering automatically. This additional variable, which is the cosine of pointing angle value for the V0 comprised of the baryon + bachelor of any given cascade candidate, is crucial for rejecting an unwanted background structure in high-multiplicity environments. More details can be found in this [analysis note](https://alice-notes.web.cern.ch/node/588).


ddobrigk:
Odd, there does not seem to be any actual error (and nothing special happening to the files I changed here?)... @dberzano, could you maybe take a quick look when you manage? Thank you!

PR #984, opened 2019-07-22
Macro to generate J/psi from B events (to be used with MCGen lego trains)

ffionda:


ffionda:
Dear all,
I want to let you know that I've tested the macro locally using MCGen_Test machinery (https://twiki.cern.ch/twiki/bin/viewauth/ALICE/AnalysisTrains#MCGen_Test ) and I've found two bugs concerning:  1) the loading of pythia6 libraries; 2) the setting of Perugia0 tune. 
Both are fixed by the last change, and the macro seems to work properly now.
Thank you

chiarazampolli:
+1

PR #986, opened 2019-07-31
AliAnalysisAlien improvements related to using the generic grid plugin interface

Atlantic777:
This pull request is required for JAliEn integration. It decouples the AliRoot analysis plugin from the legacy AliEn ROOT Grid plugin by using the interfaces that are defined in ROOT and are common for both Grid plugins.

For this purpose some new interfaces are defined in the Alice-Grid-Utils package, namely:
- [x] TAliceCollection,
- [x] TAliceFile and
- [x] TAliceJobStatus

The biggest change in this pull request is how collections are created. We no longer call `alien_find` standalone tool, but use the `gGrid->Command()` interface and build the collections in memory. The `TAliceCollection::AddFast()` method is used here exclusively for performance reasons. Use `TGridCollection` instead of `TAliceCollection` wherever possible.


pzhristov:
+1

PR #990, opened 2019-08-07
Makes TriggerTRU not contribute to L1 triggers if disabled by STU

moliver813:
Corrects TriggerTRU code so that masking by STU actually prevents the TRU from contributing to L1 Triggers.


mfasDa:
+1

PR #992, opened 2019-08-10
Skip AssignGeneratorIndex for header with particles_suppressed in the title

shahor02:
Gen.headers with "particles_suppressed" in the title will not be accounted for in the AssignGeneratorIndex.


shahor02:
+1

PR #994, opened 2019-08-13
Improve parsing the version in AliDataFile

Atlantic777:
This commit affects only builds that have multiple words in their name, such as "_NEXT_ALIEN" suffix.


pzhristov:
+1

PR #996, opened 2019-08-14
Fix a bug in the analysis plugin

Atlantic777:
Fix submitting collections for user analysis when creating collections for multiple runs.


pzhristov:
+1

PR #998, opened 2019-08-20
Sync GPU from O2

davidrohr:
@martenole : Syncs the current version and adapts the AliRoot cmake. Didn't check yet if TRD tracking works in your task, this might need more work.

But once this is merged, you can just copy over the GPU directory from O2 for your development.


davidrohr:
+test

davidrohr:
+1

PR #1005, opened 2019-08-22
Small fixes in AliTPCDcalibRes

martenole:
- fix in median calculation
- add fY2XBins* arrays to destructor
- fix GetDY2X method: before the results were not matching for the case of uniform and variable binnings. But I think this is not really an issue, since as far as I can see this function is not used

### DIFF
--- a/TPC/TPCcalib/AliTPCDcalibRes.cxx
+++ b/TPC/TPCcalib/AliTPCDcalibRes.cxx
@@ -239,6 +239,9 @@ AliTPCDcalibRes::~AliTPCDcalibRes()
   delete[] fMaxY2X;
   delete[] fDY2X;
   delete[] fDY2XI;
+  delete[] fY2XBinsCenter;
+  delete[] fY2XBinsDH;
+  delete[] fY2XBinsDI;
   delete[] fZ2XBinsCenter;
   delete[] fZ2XBinsDH;
   delete[] fZ2XBinsDI;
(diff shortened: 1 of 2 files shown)


martenole:
Hi @shahorO2
I saw that the fix for the median calculation was still not included in the master. I have added this here now. 
Also when porting the optional variable Y/X and Z/X binning I noticed some arrays missing in the destructor and a small bug in the GetDY2X method (which I think as no effect since I could not see where the function is used).
Cheers,
Ole

shahor02:
Hi @martenole 
Thanks for spotting!

shahor02:
+1

aaniin:
Hello @martenole ,
great bug spotting 🦅 !
May I ask why you stick with the pointers for the `fY2XBins`? In my naive view, you have a perfect example for not using a pointer which avoids deletion omissions right above [link](https://github.com/martenole/AliRoot/blob/7b944dd703ddb29f4ba2c4640ec2292d7f6b955b/TPC/TPCcalib/AliTPCDcalibRes.h#L500)
Do you want to teach me and elaborate your choice of using the pointer? 👨‍🏫 
Have a nice weekend! 
Hans

martenole:
Hi @aaniin 
nice to hear from you! I am pleasantly surprised that you are still following the AliRoot developments so closely :)
Actually, I am just porting this class to o2. There I avoid bare `new`/`delete`s (check the code [here](https://github.com/martenole/AliceO2/blob/tpccalib/Detectors/TPC/calibration/SpacePoints/include/SpacePoints/TrackResiduals.h)).
In `AliTPCDcalibRes` I am not really changing anything, that class has been running like this very smoothly for quite some time now.
Cheers,
Ole

PR #1012, opened 2019-09-03
Fix a bug in AliAnalysisAlien::CreateDataset

Atlantic777:
This is a proposed hotfix for a bug introduced after switching to the TGrid API for creating the Grid collections in the analysis plugin.


chiarazampolli:
+1

PR #1017, opened 2019-09-11
Support Z=2 nuclei optimised PID for tracking

mpuccio:
Hi @shahor02

this is the PR that we were mentioning before with @akalweit. This should be coupled with a PR to AliDPG to be used in production.
Do you think this is good enough?

Thanks!
Cheers,
Max


shahor02:
@mpuccio ok for the code. In principle, you don't need to create the setter in the AliReconstruction, one can directly set ``AliESDpid::SetOnly3HeOrPi(130)`` in the rec.C.
Shall I merge it as it is?

mpuccio:
@shahor02 yes, it is true, but I wanted to follow the same logic used for the Run1PIDForTracking in AliDPG (see alisw/AliDPG#307). For me it is fine to merge this, thanks a lot!

shahor02:
+1

PR #1022, opened 2019-09-16
Ensure consistent storage of Kinks in the AOD (P.Ganoti + RS)

shahor02:
Now apart from the Kink AliAODVertex also the AliAODkink object is stored, containing the
kink-specific info like kinematics of prongs at the kink point.
All mothers are stored (including those which did not get any filter bit).


shahor02:
+1

shahor02:
+1

shahor02:
+1

PR #1026, opened 2019-09-20
Replacing for Xic0 and OmegaC0 channels with ME=11 and part=81 with ME=42 and diquarks + some renormalisation to avoid warnings

arossi81:
For Xic0 and OmegaC0 the channels those with ME=11 and part=81 were raising PYTHIA errors claiming that charge was not conserve. It's not clear why the same problem does not appear for Xic+, which has identical channels (except for the spectator quarks).  I removed these channels and,  to preserve decays to several hadronic states, I added channels with ME=13 and diquarks, copying them from the current PYTHIA 8.2 decay table (online manual). The decay channels required for ITS3 simulations were not touched, a part from rescaling of some BR (also for Xic+) for removing PYTHIA warnings related to sum of BR !=100% (this just to avoid useless entries in sim.log).


ktf:
+test

pzhristov:
+1

PR #1029, opened 2019-09-23
Update acceptance cuts for ITS2 layers

strogolo:
I added data members and setters to activate acceptance cut for ITS2 simulations


shahor02:
+1

PR #1032, opened 2019-09-26
[ALIROOT-8237] Hack to set proper generator indices

miweberSMI:


miweberSMI:
Dear @shahor02, @chiarazampolli , @fprino , all,

as reported in last DPG coordination meeting we (reported by @dsekihat) have still problems in ALIROOT-8237 with the extraction of generator indices. As far as I can see the number of HIJING particles is still stored (extracted in AliGenMCEvent::AssignGeneratorIndex via AliGenEventHeader::NProduced()), but the particles are not there. Thsi leads to some improper assignment of generator indices. I found this hack to fix it (it also still works for non-embedded/filtered events), but there might be better options.
I assume that for an ESD analysis this would fix the problem, but for AODs we would have to refilter, correct?
Anyhow, let me know what you think...

Thanks and cheers,
Michael

shahor02:
Hi @miweberSMI 

When I've added {{"particles_suppressed"}} to filtered Hijing gen.header title I've tested the AOD filtering on 1 chunk and it was working. What was the problem reported at the DPG meeting?
{{fNprimaries}} corresponds to real number of stored particles (after filtering out the underlying event in non-mixed mode) in the Kinematics of the embedded event and in principle it should not be decremented by neither the {{NProduced()}} from the Hijing header nor by a constant offset 2 (did you get it empirically?)

Cheers,
 Ruben

miweberSMI:
Hi @shahor02,

thanks for your quick reply. @dsekihat reported error messages like
W-AliMCEvent::AssignGeneratorIndex: AliMCEvent::AssignGeneratorIndex: 0-pointer to particle j       -1 npart    11497 nsumpart       28 Nprimaries       28
when running over AODs (the same is true for ESDs, which I tested).
I found that the numbers in the HIJING header are very big and then get subtracted and lead to negative numbers. I am not sure I fully understand why (was the fixed not effective in this new production ALIROOT-8237?) and I just found that there is always an offset of two (so yes it is empirical).

Cheers,
Michael

shahor02:
Hi @miweberSMI 
Is there a ESD on which I can reproduce this?
Cheers,
 Ruben

miweberSMI:
Hi @shahor02 
for example: /alice/sim/2019/LHC19f2a3/297595/001/ or any other in this production...
Thanks and cheers,
Michael

shahor02:
Hi @miweberSMI 

The subtraction of NPart from the gen.header should have been done only for headers of embedded events (not flagged by ``particles_suppressed`` in the title), while it was done also for the Hijing header. My fault, sorry, looks like after testing locally the previous fix, I moved one bracket to a wrong line...

Cheers,
 Ruben

shahor02:
@miweberSMI please close this PR to avoid accidental merge.

miweberSMI:
Thanks @shahor02, I am closing this PR then...

PR #1036, opened 2019-10-04
Fix: restore AliAnalysisManager::Notify kTrueNotify actions

shahor02:
On of the changes in fix 3756604e497e65ba327a7e7b2d8f69db7b91ac09 (intended to make AliAnalysisManager compatible
both with root5.. and root6-18-04, was in fact crashing analysis with all root versions. Reverting it.


shahor02:
+1

PR #1037, opened 2019-10-04
Custom pile-up info in ESD and AOD

shahor02:
Run TPC and ITS pileup taggers from AliESDUtils in the reconstruction and fill AliESDHeader.
In the AOD filtering, either transfer pile-up info from AliESDHeader to AliAODHeader or run the taggers
directly in the filtering.


shahor02:
+1

PR #1038, opened 2019-10-08
Prefer Clear over Delete for `AliESDTRDTracklet`s

ktf:
This is where we spend 80% of the cost when iterating on an ESD collection.


ktf:
@shahor02 @pzhristov am I missing something trivial?

shahor02:
@ktf No, it should be ok to use Clear. In fact, one even does not need to use Clear("C"), just Clear() should be enough since the tracklets don't allocate memory.

ktf:
Ok, adapted to use `Clear()`.

ktf:
+test

shahor02:
+1

PR #1039, opened 2019-10-08
Clear V0s rather than deleting them

ktf:
This is now 10% of running over ESDs.


ktf:
@shahor02 same as #1038. With this applied the cost of deletion is now only for tracks (where it now accounts for ~50% of the reading time), and IIUC, that we cannot get rid of due to the non-trivial destructor, am I wrong?

ktf:
+test

shahor02:
@ktf : you are right! BTW, on the same footing one can also use ``Clear()`` instead of ``Delete()`` for
fSPDPileupVertices, fTrkPileupVertices, fPmdTracks, fCascades, fKinks, thought their number is much smaller than that for the tracks or V0's.

AliESDtrack allocates a lot of memory, in principle, one could add a ``Clear()`` method to deallocate it and  call ``fTracks->Clear("C")`` in the AliESDEvent, but the gain should be tiny.

ktf:
+1

PR #1040, opened 2019-10-10
Update galice.cuts

cepoko:
Switching on delta-ray production inside V0 scintillators.


cepoko:
Dear All,
I've tested the fix on lxplus and confirm that everything is working. Could you please approve it so that we can make an aliroot tag and launch MC production.
Thanks, Cvetan

shahor02:
+1

amorsch:
ok for me but we should check the transport time both in G3 and G4

PR #1041, opened 2019-10-10
Reset L0 time

mfasDa:
After the event is done the L0 time has to be set
to 0, otherwise it is not re-evaluated for the next
event (check for L0 time != 0)


mfasDa:
+test

pzhristov:
+1

PR #1044, opened 2019-10-17
Fix in treatment of return value of RunCustomPileUpFinders

shahor02:
Returned false should be treated as a failure, while opposite was done.

@pzhristov thanks for spotting this. Fortunately, in the productions we use ``rec.SetStopOnError(kFALSE)``, so the bug did not affect any MC.


pzhristov:
+1

PR #1050, opened 2019-10-19
PWGPP-565 - remove  \t in the string export in case csv format used

miranov25:
+ adding std map for Float_t and Double_t to be able to use in root prompt


miranov25:
+1

PR #1055, opened 2019-10-24
Optimization in PropagateToDCA

shahor02:
@fprino : In fact, there was a room for extra optimization, apart from removing duplicated cs calculation.


chiarazampolli:
+1

PR #1062, opened 2019-11-15
WIP: Update the location of AliEn Grid plugins includes dir

Atlantic777:
This update is needed to align AliRoot with updates in the Grid plugins that are required for ROOT 6.18 and MacOS Catalina.


Atlantic777:
We managed to get the Grid plugins work without this change in AliRoot. I am closing this PR.

PR #1063, opened 2019-11-17
Correct use of decayer in AliGenPythiaPlus

amorsch:
- AliPythia8 has a factory method that returns a AliDecayerPythia8 object
(implementing a method in the base class AliPythiaBase)
- this method is called by AliGenPythiaPlus to set fDecayer consistently
- modifications to the decay table can be performed via AliGenPythiaPlus ::SetForceDecay

+

some updates in pythia8.C to work with root6


chiarazampolli:
+1

PR #1065, opened 2019-11-19
Fix: float precision was leading to wrong kinematics

shahor02:
@coppedis : this was the reason of the wrong kinematics


shahor02:
+1

PR #1067, opened 2019-11-20
Fix: Stack::PushTrack must assign ntr<0 when rejecting particle

shahor02:
If the PDG is not known to the TParticlePDG then the particle is not added to the stack, nevertheless
the ntr (position of the added particle on the stack) was not set to negative value. This was signalling
to the generator which called PushTrack that the particle is added to the Stack, leading to the wrong
count of primaries in the generator header. In case the stack was empty at this moment, the following
call of KeepTrack(ntr) would lead to segm.violation, since ntr initial value is 0.
Now, the ntr is >=0 only if the particle was actually added to the stack.

ping: @ffionda @miweberSMI


shahor02:
+1

PR #1068, opened 2019-11-24
fPythia object made non-persistent to avoid default constructur being called …

amorsch:
…when reading galice.root


amorsch:
The change avoids a crash during simulation after transport when using Pythia8.

chiarazampolli:
+1

PR #1076, opened 2019-12-06
Fix index to have the CutOnChild selection working with Pythia8

fprino:
Dear @amorsch ,

with Pythia8 the HF selection on the decay particles was not working (causing in an infinite loop).
It seems to me it was due to a fortran/c++ mismatch in the counting (from 0 or from 1).
The fix in this commit solves the issue: from local simulations I could get proper selection on the decay particles.

Cheers, Francesco


chiarazampolli:
+1

fprino:
Hi,

not sure what happened here, The test were successful, but despite the +1 given by Chiara the commit was not merged.
Can you please check?

Thanks!

Ciao, Francesco

PR #1077, opened 2019-12-09
Implement kHeavyFlavor as option for SetSTackFillOpt

fprino:
HI,

this modification implements a feature that allows us to fill the stack only with particles originating from c/b quark hadronization. This feature was present in PYTHIA6 and it used in Pb-Pb D2H simulations to inject only charm (beauty) signals with PYTHIA on top of an underlying event, without injecting the "full" pp event.
The implementation is taken from AliGenPythia, with some modification to deal with the different array index (starting from 0 or from 1) in C++ and fortran.

Best, Francesco


shahor02:
+1

fprino:
Hi @ktf ,

the error from the test reported above doesn't seem related to my modifications, which affects the generator part, not the PID response. It seems an issue in accessing the OADB.
Can you please have a look?

Thanks!

Francesco

ktf:
@MichaelLettrich do you understand what is happening for this?

PR #1080, opened 2019-12-14
Fix in GetITSPileupVertexInfo: wrong array was used

shahor02:


shahor02:
+1

fprino:
Dear @ktf and @MichaelLettrich,

this PR (as the other 2 which are open since 2 days) seems stuck in build/AliRoot/macos.
The error reported above does not seem related to the modification by Ruben.
Can you please have a look?
Thanks!

Ciao, Francesco

miranov25:
Hallo Ruben.

Thank you for fix. 
It was copy/paste error. In our pile-up studies we were indeed using pileup multiplicity and  pile-up median z position - which is correct.

Marian

PR #1081, opened 2019-12-16
PWGPP-566, PWGPP-426, PWGPP-571 - TOF tracking bug fix   -using 3D distance …

miranov25:
## PWGPP-566, PWGPP-426, PWGPP-571 - TOF tracking bug fix   

Using 3D distance instead of the   localX (radial) distance
The bug was affecting PbPb productions since 2011.

Next commit (more lines to change and benchmark)  which will follow soon:
*  prohibits cluster usage (cluster SetUse) - only for "gold matches" 
  * criteria  match in space and time
  * gold to be defined in scan
*  TRD/TOF causality information for best matches


chiarazampolli:
+1

PR #1082, opened 2019-12-17
M<0 in AliV0HypSel will be interpreted as q=2 particle

shahor02:
Needed for correct treatment of the hypertriton mass evaluation.


shahor02:
+1

PR #1090, opened 2020-01-10
ATO-453- bug fix in coding hit charge for   Monopole simulation

miranov25:


miranov25:
I assume failure is a temporary problem

PR #1094, opened 2020-01-20
Updated model for fragments production in A-A collisions generated with HIJING

coppedis:
The committed changes affect ONLY the spectator production in A-A collisions when using HIJING as event generator. Therefore they are relevant only fir the ZDC.
The provided model has relative to the previously implemented one two main features:
1) it is based on the parameterization of 5 TeV data (instead of 2.76 TeV data)
2) it provides a much simpler parametrization, based only on the generated impact parameter, irrespective of the number of spectators generated by HIJING, using the experimentally measured quantities and limiting thus the model dependencies.

Ch.


chiarazampolli:
Ciao @coppedis , 

A couple of comments before approving.
I see quite many printfs, are they needed, or a leftover? (to be sure to not fill the log with unneeded information). 
And to be sure: this change does not affect anything in the underlying event, right? So it will not change the event properties with respect to existing MCs (apart from ZDC data). Will you use this for specific productions?

Cheers,

Chiara

coppedis:
Dear @chiarazampolli,
the printf have been removed.
The only change is the production of spectator nucleons in A-A collisions that is always taken from data, but has been generalized and is now based on an updated analysis of experimental data at 5.02 TeV. Apart from the spectators nothing has been changed. 

ch.

chiarazampolli:
Ok, it seems to me that this is not affecting the GenPurpose (or PWG) productions, so we should not need to port it to any tag, at least if not requested. 
Pinging @fprino for information.

Chiara

chiarazampolli:
+1

chiarazampolli:
Pinging @ktf  once more... 

Chiara

PR #1095, opened 2020-01-22
Fix remaining usage of alien_find

javierecc:
With a local installation (on a mac) of the master the merging via JDL of an analysis outputs fails. The locally created Stage_1.xml fails. As I understand the  new alien_find command does not redirect the output to the standard output and thus redirection of it to a file ends up with an empty file.


Atlantic777:
+1, this is leftovers. We should really switch to using the gGrid API everywhere.

ktf:
+test

chiarazampolli:
@ktf , also here the usual gun related error, to be ignored. We need to merge manually.

PR #1096, opened 2020-01-24
Add HF selection case with c and b proportions as in Pythia

fprino:
This feature allows to select generated events containing a charm or a beauty quark.
This allows to create samples enriched of HF with the c and b proportions given by PYTHIA, which could be practical for some studies (e.g. HF-lepton pairs).


chiarazampolli:
+1

ktf:
Errors on macOS unrelated. Merging. @chiarazampolli we should get back to the gun stuff, and eventually remove it.

PR #1097, opened 2020-01-29
Port modifications needed for HF simulations with Pythia8

fprino:
The backporting of these commits is needed to run HF simulations anchored to the pp 5 TeV runs.
Details can be found in:
https://alice.its.cern.ch/jira/browse/ALIROOT-8371
https://alice.its.cern.ch/jira/browse/ALIROOT-8396


ktf:
+test

PR #1098, opened 2020-01-30
Ignore ZMQ when Aliroot is built with root5

shahor02:
Root5 fails to generate a dictionary requiring ZMQ (v4.2 in my case)

ping to @ktf @mkrzewic

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,7 +317,9 @@ else()
   find_package(FASTJET)
 
   # ZEROMQ
-  find_package(ZeroMQ)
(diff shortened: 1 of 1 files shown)


mkrzewic:
this used to work without problems with root5 - what changed? What dictionary generation is failing?

davidrohr:
ROOT5 CINT is unable to understand the zmq.h, depending on zmq version.
I think it is safe to disable this.

mkrzewic:
indeed, seems safe to go this way.

shahor02:
+1

chiarazampolli:
@ktf , can you do your magic? :-)

PR #1100, opened 2020-01-30
Human-readable version of the AliEMCALTriggerData

mfasDa:
Previous interface was written in a completely cryptic
way and ununderstanable for non-experts. Single-letter
variable names were meaningless and do not explain their
purpose. Also from the function names do not expose
their purpose. Allthough the logics is correct this makes
the class hard to use for non-experts

In the new version parameter names were changed. Wherever
the value space corresponds to certain states enumeration
values are used. The new interfaces clearly explain their
purpose.

In addition doxygen comments were added for each function
where the purpose could be extracted easily. With the
new interfaces and the doxygen comments the class should
be self-explaining also for non-expert users.


mfasDa:
+test

PR #1103, opened 2020-02-03
PWGPP-571, PWGPP-529 -TStatToolkit::MakePDFMap()

miranov25:
## More options needed for the NDimensional pipeline
This function replace previously used TStatToolkit::MakeDistortionMapFast
Options specified like in Python version - using map-dictionary
* option to sore histograms
* option to store cumulative
* ... more options - statistic export  to be added ...


miranov25:
+1

miranov25:
+1

ktf:
The issue seems to be a genuine one. I think you are missing an `#include <string>`.

miranov25:
+1
For me the code is compiling and working properly  on my laptop and at GSI farm. 
Looks like different compiler. Strange in both cases I assumed we build compiler ...

miranov25:
looks like problem with server
@ktf 
could you check?

PR #1105, opened 2020-02-05
Add streaming operators for EMCAL raw digits

mfasDa:
In addition improved documentation that actually hints
the meaning of the functions.


mfasDa:
+1

PR #1106, opened 2020-02-05
Missing settings for altro clock fix

shahor02:
Previous patch was working for clusters produced in the reconstruction, but was failing
when clusters were extracted beforehand


chiarazampolli:
Hello @ktf , 

Can you please merge? Then I will cherry pick this to v5-09-53-patches, and we will need a new tag.

Chiara

PR #1110, opened 2020-02-07
Avoid double defined sections in Pcon.

amorsch:
Avoid double defined sections in Pcon. 
Created warnings during geometry creation.


chiarazampolli:
+1

PR #1115, opened 2020-02-16
Overwrite default process (elastic scattering) from decayer initialis…

amorsch:
Overwrite default process (elastic scattering) from decayer initialisation (solves ALIROOT-8407)


chiarazampolli:
+1

PR #1116, opened 2020-02-17
Add option to trigger particles over any of the central barrel calos: PHOS or EMCal or DCal

gconesab:
Option to avoid 2 cycles in MC productions like until now, one for EMCal and another one for DCal+PHOS for gamma-jet or jet-jet triggered on gamma over calorimeter acceptance.
With these changes, it is possible to do the OR on the acceptance of the 3 calorimeters so that one MC cycle can be requested, and for better event weight scaling between calorimeters.


chiarazampolli:
+test

chiarazampolli:
+1

PR #1128, opened 2020-03-16
switch ON pi0 decay in pythia8 decayer

amorsch:
Since AliDecayerPythia8 is correctly called from AliPythia8 the decay of pions was switched off.
With this change the pi0 decays are ON again.


chiarazampolli:
+1

PR #1130, opened 2020-03-19
ALIROOT-8420- set lambda OCDB as used first time for the LHC18q,r

miranov25:
patch obtained from lambda OCDB as described in the ALIROOT-8420


miranov25:
+1

PR #1141, opened 2020-04-07
Elastic scattering between N-N pairs switched off to avoid frequent crashes

amorsch:
The crashes are due to a numerical instabilities in the elastic scattering routine at specific settings of the center of mass energy. In these cases for some nucleons E < |pz|. This is checked by HIJING and the program stops.
No significant effects from elastic scattering on physics expected at LHC energies.
With 
hijing->SetNoElas(kFALSE);   
one can switch elastic scattering on.


chiarazampolli:
+1

PR #1143, opened 2020-04-09
Change default value for labels from 0 to -1, since 0 is a valid value

amorsch:
Requested by @loizides


chiarazampolli:
+1

chiarazampolli:
@amorsch , @fprino , do we need to inform the analyzers about this change? Probably not. It looks to me a sort of fix.

PR #1147, opened 2020-04-22
[EMCAL 510] Fix TRU L0 version selection

moliver813:
[EMCAL 510] The TRU firmware version was not saved properly and should not
be used for determine which L0 version to use. Decision is now made based
on the L0 Sel variable, via AliEMCALTriggerTRUDCSConfig::GetSegmentation().
This corrects issue where simulations used L0v0 instead of L0v1


gconesab:
+test

gconesab:
+1

gconesab:
hi @ktf 

macos testing seems stuck, maybe for the same reason you announced by mail?

ktf:
@MichaelLettrich can you check? In principle it should not matter for macOS, as we do not use the same setup.

MichaelLettrich:
@ktf  @gconesab hi, for some reason the buldscript got terminated so I started it again.

gconesab:
+1

PR #1150, opened 2020-04-28
PushTrack(..., Float* pmom ...): calculate energy in double precision

amorsch:
Should solve 
https://alice.its.cern.ch/jira/browse/ALIROOT-8233

AliGenHijing uses the AliStack::PushTrack(Int_t done, Int_t parent, Int_t pdg, const Float_t *pmo
method of AliStack
This method calculates the energy from the momentum and the particle mass.
The calculation was single precision and has been changed to double precision now.


sawenzel:
+1

PR #1154, opened 2020-05-06
PushTrack(..., Float* pmom ...): calculate energy in double precision

gconesab:
Needed for ALIROOT-8423, ALIROOT-8425


chiarazampolli:
Thanks @gconesab !

PR #1156, opened 2020-05-16
FIx sign of Xic+ decay products

fprino:


chiarazampolli:
Ciao @fprino , 

Before merging: Andrea Dubla mentioned in JIRA (ALIROOT-8281) that he wanted to first check if other channels were also affected. Do you think that we should wait, in case more fixes come, or merge already now? 

Cheers,

Chiara

chiarazampolli:
Merging for now, if needed, more commits will be added.

PR #1157, opened 2020-05-26
Provide clear specification for function parameters

mfasDa:
Provide specification consisting of parameter explanation,
use case, and supported value range to getter / setter
functions.


mfasDa:
+1

Affecting only doxygen comment. These are absolutely essential for the usage. Without parameter names, meaning and ranges (corresponding values) are not clear, i.e. one needs to check in the OCDB in order to figure out what index corresponds to which threshold for the threshold values).

mfasDa:
+1

PR #1167, opened 2020-06-22
User trigger implementation to select generated events 

lucamicheletti93:
Implementation of a trigger function to select generated events. The method can be set directly by the user from the lego train interface and it can be used with external generators (e.g. JEWEL) or with AliRoot internal generators (e.g. PYTHIA).

### DIFF
--- a/EVGEN/AliGenExtFile.cxx
+++ b/EVGEN/AliGenExtFile.cxx
@@ -31,10 +31,6 @@
 
 #include <Riostream.h>
 
-///
-#include <functional>
-///
-
 #include "AliLog.h"
 #include "AliGenExtFile.h"
 #include "AliRunLoader.h"
@@ -294,46 +290,30 @@ void AliGenExtFile::Generate()
 	fNprimaries++;
     } // track loop
 
-    ///
-    printf("--------------------------- \n");
     // User Trigger
     if(fSetUserTrig){
-      printf("---> USER TRIGGER ENABLED \n");
       AliStack *stack = AliRunLoader::Instance()->Stack();
-      if(!fUserTrigger(stack)){
-        printf("EVENT DISCARDED \n");
-        continue;
-      }
+      if(!fUserTrigger(stack)){continue;}
     }
-    else{printf("---> USER TRIGGER DO NOT SET \n");}
 
     // Base multiplicity trigger
     if(fSetMultTrig){
-      printf("---> MULTIPLICITY TRIGGER ENABLED \n");
-      printf("mult cut : %i \n",fMultCut);
+      AliInfo(Form("mult cut : %i",fMultCut));
       AliStack *stack = AliRunLoader::Instance()->Stack();
       if(!MultiplicityTrigger(stack)){
         consecutiveDiscardedEvents++;
         if(consecutiveDiscardedEvents>10){AliFatal("More than 10 events discarded consequently");}
-        printf("EVENT DISCARDED \n");
         continue;
       }
       consecutiveDiscardedEvents=0;
     }
-    else{printf("---> MULTIPLICITY TRIGGER DO NOT SET \n");}
 
     // Base pT trigger
     if(fSetPtTrig){
-      printf("---> PT TRIGGER ENABLED \n");
-      printf("pT cut : %f GeV/c \n",fPtCut);
+      AliInfo(Form("pT cut : %f GeV/c",fPtCut));
       AliStack *stack = AliRunLoader::Instance()->Stack();
-      if(!PtTrigger(stack)){
-        printf("EVENT DISCARDED \n");
-        continue;
-      }
+      if(!PtTrigger(stack)){continue;}
     }
-    else{printf("---> PT TRIGGER DO NOT SET \n");}
-    printf("--------------------------- \n");
     ///
 
     // Generated event header
@@ -361,10 +341,10 @@ void AliGenExtFile::CdEventFile()
 //__________________________________________________________
 Bool_t AliGenExtFile::MultiplicityTrigger(AliStack *stack){
   Int_t nTracks  = stack->GetNtrack();
-  printf("n Tracks = %i \n",nTracks);
+  AliInfo(Form("n Tracks = %i",nTracks));
   if(nTracks>fMultCut){return kTRUE;}
   else{
-    printf("n Tracks < %i --> EVENT DISCARDED \n",fMultCut);
+    AliInfo(Form("n Tracks < %i --> EVENT DISCARDED",fMultCut));
     return kFALSE;
   }
 }
@@ -376,14 +356,14 @@ Bool_t AliGenExtFile::PtTrigger(AliStack *stack){
   for(int i = 0;i < nTracks;i++){
     TParticle *track = (TParticle*) stack->Particle(i);
     if(track->Pt() > fPtCut){
-      printf("pT = %f \n",track->Pt());
+      AliInfo(Form("pT = %f",track->Pt()));
       partCounter++;
     }
   }
 
   if(partCounter>0){return kTRUE;}
   else{
-    printf("No particle with pT > %f GeV/c --> EVENT DISCARDED \n",fPtCut);
+    AliInfo(Form("No particle with pT > %f GeV/c --> EVENT DISCARDED",fPtCut));
     return kFALSE;
   }
 }
--- a/EVGEN/AliGenExtFile.cxx
+++ b/EVGEN/AliGenExtFile.cxx
@@ -56,8 +56,11 @@ AliGenExtFile::AliGenExtFile()
      fReader(0),
      fStartEvent(0),
      ///
+     fLimitDiscardedEvents(999999),
      fSetMultTrig(0),
+     fMultCut(0),
      fSetPtTrig(0),
+     fPtCut(0),
      fSetUserTrig(0)
      ///
 {
@@ -73,8 +76,11 @@ AliGenExtFile::AliGenExtFile(Int_t npart)
      fReader(0),
      fStartEvent(0),
      ///
+     fLimitDiscardedEvents(999999),
      fSetMultTrig(0),
+     fMultCut(0),
      fSetPtTrig(0),
+     fPtCut(0),
      fSetUserTrig(0)
      ///
 {
@@ -293,7 +299,12 @@ void AliGenExtFile::Generate()
     // User Trigger
     if(fSetUserTrig){
       AliStack *stack = AliRunLoader::Instance()->Stack();
-      if(!fUserTrigger(stack)){continue;}
+      if(!fUserTrigger(stack)){
+        consecutiveDiscardedEvents++;
+        if(consecutiveDiscardedEvents>fLimitDiscardedEvents){AliFatal(Form("More than %i events discarded consequently",fLimitDiscardedEvents));}
+        continue;
+      }
+      consecutiveDiscardedEvents=0;
     }
 
     // Base multiplicity trigger
@@ -302,7 +313,7 @@ void AliGenExtFile::Generate()
       AliStack *stack = AliRunLoader::Instance()->Stack();
       if(!MultiplicityTrigger(stack)){
         consecutiveDiscardedEvents++;
-        if(consecutiveDiscardedEvents>10){AliFatal("More than 10 events discarded consequently");}
+        if(consecutiveDiscardedEvents>fLimitDiscardedEvents){AliFatal(Form("More than %i events discarded consequently",fLimitDiscardedEvents));}
         continue;
       }
       consecutiveDiscardedEvents=0;
@@ -312,7 +323,12 @@ void AliGenExtFile::Generate()
     if(fSetPtTrig){
       AliInfo(Form("pT cut : %f GeV/c",fPtCut));
       AliStack *stack = AliRunLoader::Instance()->Stack();
-      if(!PtTrigger(stack)){continue;}
+      if(!PtTrigger(stack)){
+        consecutiveDiscardedEvents++;
+        if(consecutiveDiscardedEvents>fLimitDiscardedEvents){AliFatal(Form("More than %i events discarded consequently",fLimitDiscardedEvents));}
+        continue;
+      }
+      consecutiveDiscardedEvents=0;
     }
     ///
 
--- a/EVGEN/AliGenExtFile.cxx
+++ b/EVGEN/AliGenExtFile.cxx
@@ -31,6 +31,10 @@
 
 #include <Riostream.h>
 
+///
+#include <functional>
+///
+
 #include "AliLog.h"
 #include "AliGenExtFile.h"
 #include "AliRunLoader.h"
@@ -54,7 +58,12 @@ AliGenExtFile::AliGenExtFile()
     :AliGenMC(),
      fFileName(0),
      fReader(0),
-     fStartEvent(0)
+     fStartEvent(0),
+     ///
+     fSetMultTrig(0),
+     fSetPtTrig(0),
+     fSetUserTrig(0)
+     ///
 {
 //  Constructor
 //
@@ -66,7 +75,12 @@ AliGenExtFile::AliGenExtFile(Int_t npart)
     :AliGenMC(npart),
      fFileName(0),
      fReader(0),
-     fStartEvent(0)
+     fStartEvent(0),
+     ///
+     fSetMultTrig(0),
+     fSetPtTrig(0),
+     fSetUserTrig(0)
+     ///
 {
 //  Constructor
     fName   = "ExtFile";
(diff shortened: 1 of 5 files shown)


EVGEN/AliGenExtFile.cxx
```diff
@@ -276,6 +290,32 @@ void AliGenExtFile::Generate()
...
+      if(!fUserTrigger(stack)){continue;}
+    }
+
+    // Base multiplicity trigger
+    if(fSetMultTrig){
+      AliInfo(Form("mult cut : %i",fMultCut));
+      AliStack *stack = AliRunLoader::Instance()->Stack();
+      if(!MultiplicityTrigger(stack)){
+        consecutiveDiscardedEvents++;
+        if(consecutiveDiscardedEvents>10){AliFatal("More than 10 events discarded consequently");}
```
gconesab, line 305: Why is it needed to count how many events are discarded and not allow more than 10? I guess if you request a very large multiplicity in pp events it is possible to have such situation

EVGEN/AliGenExtFile.cxx
```diff
@@ -298,8 +338,32 @@ void AliGenExtFile::CdEventFile()
 // CD back to the event file
   AliRunLoader::Instance()->CdGAFile();
 }
+//__________________________________________________________
+Bool_t AliGenExtFile::MultiplicityTrigger(AliStack *stack){
+  Int_t nTracks  = stack->GetNtrack();
+  AliInfo(Form("n Tracks = %i",nTracks));
+  if(nTracks>fMultCut){return kTRUE;}
```
gconesab, line 345: Would it not be interesting to apply not only a high limit but also a low limit?

PYTHIA6/AliPythia6/AliGenPythia.cxx
```diff
@@ -54,9 +54,9 @@ ClassImp(AliGenPythia)
 
 AliGenPythia::AliGenPythia():
     AliGenMC(),
-    fProcess(kPyCharm),          
+    fProcess(kPyCharm),
```
gconesab, line 57: In this class, there are many diff lines like this, where no real intervention happened. It would be good to avoid this, although I do not know how to redo the commit

EVGEN/AliGenExtFile.cxx
lucamicheletti93: Indeed you are right! I added this line when I was testing different multiplicity cuts in order to avoid runnung even if no event is found. Are you siggesting to remove it or to simply increase the threshold for discarded events?

EVGEN/AliGenExtFile.cxx
lucamicheletti93: AliGenExtFile::MultiplicityTrigger() and AliGenExtFile::PtTrigger() are mainly meant to be two easy-to-use methods which can also provide an example for the user to implement his/her own trigger with fUserTrigger(stack)

EVGEN/AliGenExtFile.cxx
gconesab: At least not a hardcoded number. Devote a dedicated data member and put as default a large number (99999?) and a print in case this number is reached.

EVGEN/AliGenExtFile.cxx
gconesab: OK

EVGEN/AliGenExtFile.cxx
lucamicheletti93: Thanks! I have addd the data member as you suggested in my new commit.

PYTHIA6/AliPythia6/AliGenPythia.cxx
lucamicheletti93: I have tried to recover the master version re-adding my modification but nothing has chenged...

gconesab:
It looks good to me, I let the superpowers to test/approve unless any other comments appear.

ktf:
+1

chiarazampolli:
The error does not look like related to the PR, unless I overlooked something... @ktf , can you take a look?

STEER/STEER/AliGenerator.cxx
```diff
@@ -450,19 +450,33 @@ void AliGenerator:: SetHighWaterMark(Int_t nt)
...
-   
+
 }
 void AliGenerator::FinishRun()
 {
     ;
 }
 
+///-------------------------------------------------------------------------
+Bool_t AliGenerator:: ApplyUserTrigger()
```
chiarazampolli, line 463: @lucamicheletti93 : please check teh return value, in some branches of your conditions, you do not return anything.

chiarazampolli:
Actually there is an error, @lucamicheletti93 , can you take a look at my comment?

STEER/STEER/AliGenerator.cxx
lucamicheletti93: I think the issue is due to that if ApplyUserTrigger() is called but fSetUserTrig == kFALSE nothing is returned. I should add a return kTRUE in case fSetUserTrig == kFALSE.

chiarazampolli: Yes. You should be sure that something is always returned.

lucamicheletti93: I re-commit the code after applying this correction!

STEER/STEER/AliGenerator.cxx
```diff
@@ -450,19 +450,37 @@ void AliGenerator:: SetHighWaterMark(Int_t nt)
...
+    if(!fUserTrigger){
+      AliFatal("No external trigger defined!");
+      return kTRUE;
+    }
+    else{
+      if(!fUserTrigger(fStack)){return kFALSE;}
+      else{return kTRUE;}
+    }
+  }
+  else{return kTRUE;}
```
chiarazampolli, line 476: you don't need the last "else", and neither the parenthesis after it (on line 476 nbut also 473), if it is only one line.

lucamicheletti93: Perfect! I remove the lase else and I re-commit

STEER/STEER/AliGenerator.cxx
```diff
@@ -450,19 +450,36 @@ void AliGenerator:: SetHighWaterMark(Int_t nt)
...
+    if(!fUserTrigger){
+      AliFatal("No external trigger defined!");
+      return kTRUE;
+    }
+    else{
+      if(!fUserTrigger(fStack)){return kFALSE;}
+      else return kTRUE;
+    }
+  }
+}
```
chiarazampolli, line 476: Ciao Luca,
Sorry, my messages were not clear probably, I think that you removed again the last return... This is needed. Just don't put "else return kFALSE" (or kTRUE, i don't remember), but simply "return kFALS" (or kTRUE).

lucamicheletti93: Sorry, I misunderstood! Now it is fixed

chiarazampolli: Thanks!

chiarazampolli:
@lucamicheletti93 , can you squash the latest commits?

chiarazampolli:
+test

chiarazampolli:
@ktf , @shahor02 , can you also do a "+test"?

shahor02:
+test

STEER/STEER/AliGenerator.h
```diff
@@ -89,28 +93,32 @@ class AliGenerator : public TNamed, public AliRndm
...
 	{fProjectile = proj; fAProjectile = a; fZProjectile = z;}
     virtual void    SetTarget(TString tar="", Int_t a = 0, Int_t z = 0)
 	{fTarget = tar; fATarget = a; fZTarget = z;}
-    
+
     virtual void    SetSeed(UInt_t seed);
-    
+
+    ///-------------------------------------------------------------------------
+    virtual void SetUserTrigger(std::function<Bool_t(AliStack*)> fn) { fSetUserTrig = kTRUE; fUserTrigger = fn; }     // Enable user trigger
```
chiarazampolli, line 119: Ciao @lucamicheletti93 , 
You need to add the include of AliStack.h and "#include <functional>", or it won't compile.

lucamicheletti93: I have added the missing include in AliGenerator.h  and I have committed the code!

chiarazampolli:
+test

chiarazampolli:
+test

chiarazampolli:
@ktf , @shahor02 , can you also “+test” it?

shahor02:
+test

chiarazampolli:
Hello @ktf @shahor02 ,

I am not sure what is happening here... It complains:

- macos --> include "AliStack.h" 
```
/System/Volumes/Data/build/ci_checks/aliroot_checker_macos_next-root6_ci_0/sw/SOURCES/AliRoot/1167/0/STEER/STEER/AliGenerator.h:20:10: fatal error: 'AliStack.h' file not found
#include "AliStack.h"
```
- release --> it does not recognize std::function and AliStack, even if the includes are there (and actually, I notice now that AliStack was declared already as "class")
```
[ 12%] Building Fortran object MICROCERN/CMakeFiles/microcern-object.dir/kerset.F.o
Error: Symbol function is not defined in current scope  /mnt/mesos/sandbox/sandbox/sw/SOURCES/AliRoot/build_AliRoot_release/0/STEER/STEER/AliGenerator.h:120:
Error: Symbol AliStack is not defined in current scope  /mnt/mesos/sandbox/sandbox/sw/SOURCES/AliRoot/build_AliRoot_release/0/STEER/STEER/AliGenerator.h:120:
Syntax Error: AliStack* /mnt/mesos/sandbox/sandbox/sw/SOURCES/AliRoot/build_AliRoot_release/0/STEER/STEER/AliGenerator.h:120:
Error: Symbol fn is not defined in current scope  /mnt/mesos/sandbox/sandbox/sw/SOURCES/AliRoot/build_AliRoot_release/0/STEER/STEER/AliGenerator.h:120:
Error: void type variable can not be declared /mnt/mesos/sandbox/sandbox/sw/SOURCES/AliRoot/build_AliRoot_release/0/STEER/STEER/AliGenerator.h:127:
```

@lucamicheletti93 : how are you compiling? To you it works, doesn't it?

Chiara

lucamicheletti93:
Dear @chiarazampolli 

> @lucamicheletti93 : how are you compiling? To you it works, doesn't it?

Locally it works for me. I'm using aliBuild build AliPhysics --defaults user-next-root6.

shahor02:
The problem appears when generating the root dictionary for the AliGenerator: the AliStack.h, being from another directory, is not appearing in the list of headers provided to rootcint. Can you try to add ``#include Stack.h`` to STEERLinkDef.h ?

chiarazampolli:
I was thinking that being AliGenerator in STEER and AliStack in STEERBase, they would have seen each other. And indeed there is the STEERBase in the CMake file https://github.com/alisw/AliRoot/blob/master/STEER/STEER/CMakeLists.txt.
And why then does it work for Luca?

shahor02:
neither rootcint itself nor the makefile resolver the dependency of the dictionaries to create on different headers, rootcint takes whatever is provided on the command line or via include dirs. On the command line the headers of the same directory are provided.
I suspect it works for Luca locally with incremental make with alienv already initialized and the root include_dir containing AliStack.
Luca, does it work also for the clean aliBuild from scratch?

chiarazampolli:
How does it work here:
https://github.com/alisw/AliRoot/blob/9e5bb12171e8782c17446eb0faf6a4b40b78c9bb/EVGEN/AliGenCocktailAfterBurner.h
?

shahor02:
@chiarazampolli works what? AliGenCocktailAfterBurner.h does not include the AliStack (just fwd declaration as it was in the AliGenerator).

chiarazampolli:
You are right. 
Never the less, in the previous version there was no include, but just the forward declaration, which I had not noticed until later. And still we got the same error in the release compilation (I guess SL6), not on macos (where there were errors related to G4). In the release compilation, it also complained about std::function. 
Now the situation is that neither release, nor macos work:

- release: it complains about std::function (but we have the include now), and AliStack (we can remove the include, and leave the forward declaration as before)
- macos: AliStack.

So, we can remove the "include AliStack", leave the "class AliStack", but if it behaves as yesterday, release will still not work, and macos will give issues in G4. @lucamicheletti93 , can you try?

BTW, any idea about why std::function with the include of "functional" does not work?

Cheers,

Chiara

lucamicheletti93:
Dear @chiarazampolli,
as you suggested I have removed the include from AliGenerator.h and I have recommitted the code!

chiarazampolli:
+test

chiarazampolli:
@shahor02 , @ktf , could you please +test this?

shahor02:
For me it fails also locally, with or w/o include. I see that rootcint generates for it a function which is not recognized by root, this might be rootcint limitation to use std::function. I will play bit more but perhaps it will be easier to use simple function pointer.

shahor02:
+test

shahor02:
Hi @lucamicheletti93 

Locally (with root 5) I always get 
````
/home/shahoian/alice/sw/BUILD/AliRoot-latest/AliRoot/STEER/STEER/G__STEER.cxx: In function ‘ROOT::TGenericClassInfo* ROOTDict::GenerateInitInstanceLocal(const AliGenerator*)’:
/home/shahoian/alice/sw/BUILD/AliRoot-latest/AliRoot/STEER/STEER/G__STEER.cxx:589:42: error: no matching function for call to ‘ROOT::TGenericClassInfo::TGenericClassInfo(const char [13], const char [76], int, const std::type_info&, const ROOT::TInitBehavior*, void (*)(), TVirtualIsAProxy*&, int, long unsigned int)’
                   sizeof(::AliGenerator) );
````
i.e. rootcint generates a call for non-existent constructor of TGenericClassInfo. To me it looks like a rootcint bug. Can you change the code to use a function pointer?

**Comment by lucamicheletti93** [comment]
Dear @shahor02,
I always use root6 for the testing of the code and it neve gave me this kind of issue, so probably using std::function with root5 could cause the problem. I try to re-implement the method using the pointer to a function (bool (*userTrigger) (AliStack)) as you are suggesting! 

Thank you very much,
Luca

**Comment by shahor02** [comment]
+test

File: STEER/STEER/AliGenerator.h [hunk 8]
```diff
@@ -89,28 +89,32 @@ class AliGenerator : public TNamed, public AliRndm
...
 	{fProjectile = proj; fAProjectile = a; fZProjectile = z;}
     virtual void    SetTarget(TString tar="", Int_t a = 0, Int_t z = 0)
 	{fTarget = tar; fATarget = a; fZTarget = z;}
-    
+
     virtual void    SetSeed(UInt_t seed);
-    
+
+    ///-------------------------------------------------------------------------
+    virtual void SetUserTrigger(Bool_t (*fn)(AliStack*)) { fSetUserTrig = kTRUE; fUserTrigger = fn; }     // Enable user trigger
```
chiarazampolli, line 115: Hello @lucamicheletti93 , @shahor02 ,

Shouldn't variables starting with "f" be private members of the class in AliRoot? Or it is only that all private members start with "f", but not all variables that start with "f" are private members?

Chiara

chiarazampolli:
The error is now in:
Processing simG4.C(1,"/mnt/mesos/sandbox/sandbox/sw/slc7_x86-64/AliRoot/build_AliRoot_release-1/test/vmctest/gun/g4Config.C"):
```
-------- EEEE ------- G4Exception-START -------- EEEE -------
*** G4Exception : PART70000
      issued by : G4NuclideTable
G4ENSDFSTATEDATA environment variable must be set
*** Fatal Exception *** core dump ***

 *** Break *** segmentation violation
Function simG4() busy flag cleared
Function Config() busy flag cleared
```
Who is responsible for this part?

Anyway, it does not look related to Luca's changes. If macos passes teh test or shows the same, I would merge. @shahor02 , what do you think?
Note also my last comment to the PR, though.

Chiara

STEER/STEER/AliGenerator.h
shahor02: In aliroot the data members start with fCapitalized, but there is not policy on function argument names.

chiarazampolli: Ok, thanks for confirming!

STEER/STEER/AliGenerator.h
```diff
@@ -179,6 +187,12 @@ class AliGenerator : public TNamed, public AliRndm
     TString      fProjectile;    // Projectile
     TString      fTarget;        // Target
 
+    ///-------------------------------------------------------------------------
+    // Custom trigger
+    Bool_t fSetUserTrig;
+    Bool_t (*fUserTrigger)(AliStack*);
```
shahor02, line 193: @lucamicheletti93 please init fSetUserTrig and fUserTrigger to false and 0 resp. in the constructors

chiarazampolli:
Same failure in https://github.com/alisw/AliRoot/pull/1170

STEER/STEER/AliGenerator.h
lucamicheletti93: Included In the last commit!

shahor02:
+test

chiarazampolli:
Same issue with G4 as before.. 
let’s wait for macos.

Chiara

shahor02:
G4 failures are unrelated

shahor02:
... but since the tests are not passed, only @ktf can merge.

sawenzel:
Something must have changed in Geant4 or our environment lately. I see the same Geant4 problem in O2 sometimes.

chiarazampolli:
Thanks to all! I tried to add a comment earlier, but it did not work.

PR #1189, opened 2020-08-04
Fix in initialization of last particle in the stack

fprino:
This modification is needed for the separate simulation of trigger event


chiarazampolli:
+1

PR #1197, opened 2020-08-27
Set generation time in cocktail header 

fprino:
This modification is needed to have the proper production time for signals injected with embedding in case of simulations with pileup (see ALIROOT-8525)

### DIFF
--- a/EVGEN/AliGenCocktail.cxx
+++ b/EVGEN/AliGenCocktail.cxx
@@ -260,6 +260,8 @@ void AliGenCocktail::Generate()
           ntimes = nsig;
         }
       }
+      AliInfo(Form("Generator %d: %s; ntimes=%d; Vertex (%f,%f,%f) cm   Time %f ns",
+		   igen,gen->ClassName(),ntimes,fVertex.At(0), fVertex.At(1), fVertex.At(2), fTime*1e9));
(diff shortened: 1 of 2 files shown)


fprino:
Hi,

this modification fixes the problem reported in ALIROOT-8525.
In case of productions with signal embedding on top of pileup events there was a mismatch betwee the production time of the injected signals and those of the underlying event.
The injected signals were assigned a production time = 0, while the background event (the trigger one) had a time distributed according to the bunch crossing time.
So, there was a difference of some tens of ps (reaching up to 400 ps) between the production time of signals and background.
This caused issues in the TOF algorithm for the determination of the start time.

Wit this modification the time of the collision is stored in the header of the cocktail and at the moment of the signal generation for the embedding it is retrieved in AliVertexGenFile::GetVertex() and assigned to the signals.

Local tests are successful.

@shahor02: can you please have a look that these changes are OK and do not have side effects?

Thanks!

Cheers, Francesco

shahor02 (approved):
Hi @fprino 
Correct, the trigger vertex time should be passed via global header.

shahor02:
+1

fprino:
Hi Ruben,

thanks for checking and approving.

Cheers, Francesco

PR #1209, opened 2020-09-02
Generate new vertex position for trigger event for embedding into pileup bkg

fprino:
This commit is needed to have proper generation of the primary vertex (position and time) for productions with MC-to-MC embedding in which we will use AliGenPileup to generate the pileup events as background and the embedding to plug the trigger event as the signal on top of the pileup.
This strategy is the one that we plan to use for the productions anchored to the p-Pb periods with high interaction rate.


chiarazampolli:
+1

fprino:
Dear @ktf and @MichaelLettrich,

it seems we have an issue with the build/AliRoot/macos test.
Can you please have a look?

Many thanks!
Best regards,
                           Francesco

MichaelLettrich:
@fprino looking into it. furthermore macos machine is down for system update currently. Should be back before 14:00 today.

PR #1213, opened 2020-09-04
Add option for semileptonic decay of Xic0

fgrosa:
Add option to force Xic0 -> Xi- e+ nu_e decay instead of Xic0 -> Xi- pi+ decay, needed for D2H MC productions

PR #1214, opened 2020-09-09
Add decay channels of Xic and Omegac not present in pythia8 decay table

fgrosa:
@chiarazampolli in this PR https://github.com/alisw/AliRoot/pull/1213 I forgot to add the Xic decay channel, which I found out not to be in the pythia8 decay table. I added it with this PR, together with the Omegac decay which was also missing. Thanks!


chiarazampolli:
+1

PR #1222, opened 2020-09-15
Added function in AliGenHijing to change Pz sign

rihanphys:
Added a function in AliGenHijing to (optionally) change Pz sign.

### DIFF
--- a/THijing/AliGenHijing.cxx
+++ b/THijing/AliGenHijing.cxx
@@ -89,7 +89,8 @@ AliGenHijing::AliGenHijing()
      fFragmNWidth(0x0),
      fFragmPWidth(0x0),
      fEConv(1),
-     fMissing(50)
+     fMissing(50),
+     fFlipPzSign(kFALSE)
 {
   // Constructor
   fEnergyCMS = 5500.;
@@ -146,7 +147,8 @@ AliGenHijing::AliGenHijing(Int_t npart)
      fFragmNWidth(0x0),
      fFragmPWidth(0x0),
      fEConv(1),
-     fMissing(50)
+     fMissing(50),
+     fFlipPzSign(kFALSE)     
 {
 // Default PbPb collisions at 5. 5 TeV
 //
@@ -448,6 +450,12 @@ void AliGenHijing::Generate()
 	      origin[2] = origin0[2]+iparticle->Vz()/10;
 	      tof = time0+kconv * iparticle->T();
 
+	      if(fFlipPzSign){
+		p[2] = -p[2];
+		origin[2] = origin0[2] - iparticle->Vz()/10;
+	      }
+
+	      
 	      imo = -1;
 	      TParticle* mother = 0;
 	      if (hasMother) {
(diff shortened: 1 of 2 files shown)


rihanphys:
I added a  function SetFlipTrackPz(Bool_t b) with Boolean argument in the AliGenHijing class to (optionally) change the sign of Pz of the primary tracks. This is needed to simulate fixed target collisions for the ALICE Fixed Target program. In the LAB frame, the projectile is moving in the +ve Z direction while target is fixed. This causes the produced particle to be boosted in the forward rapidity (+ve eta).  However, for the fixed target simulation in ALICE, we place the target at +480m and the projectile should be moving in -ve Z direction. So that the particle are boosted in the -ve eta region where we would have our tracking detectors (in the MC simulation). The simplest way to obtain this is to change the direction of Pz for primary tracks. Hence the function is added. 

Important Note : By default the Boolean is set to kFALSE, hence the track Pz sign is NOT changed. Therefore this change does not affect any existing/or future analysis with the class in its nominal settings.

rihanphys:
I added a  function SetFlipTrackPz(Bool_t b) with Boolean argument in the AliGenHijing class to (optionally) change the sign of Pz of the primary tracks. This is needed to simulate fixed target collisions for the ALICE Fixed Target program. In the LAB frame, the projectile is moving in the +ve Z direction while target is fixed. This causes the produced particle to be boosted in the forward rapidity (+ve eta).  However, for the fixed target simulation in ALICE, we place the target at +480m and the projectile should be moving in -ve Z direction. So that the particle are boosted in the -ve eta region where we would have our tracking detectors (in the MC simulation). The simplest way to obtain this is to change the direction of Pz for primary tracks. Hence the function is added. 

Important Note : By default the Boolean is set to kFALSE, hence the track Pz sign is NOT changed. Therefore this change does not affect any existing/or future analysis with the class in its nominal settings.

chiarazampolli:
Looks ok to me. @amorsch , could you also check?

Thanks!

Chiara

amorsch:
> Looks ok to me. @amorsch , could you also check?
> 
> Thanks!
> 
> Chiara

Indeed, looks ok.
Cheers
Andreas

chiarazampolli:
+1

chiarazampolli:
Thanks @amorsch !
@rihanphys , note that to use this on the grid, or with CVMFS, you need a new AliRoot tag. Is it urgent? 

Chiara

rihanphys:
@chiarazampolli, I managed to run the simulation for run3 with the changes to AliGenHijing locally in my laptop. Therefore, it is not so urgent to have a new AliRoot tag.  However, I see some error in the merging. I have to check if that is originating from my changes.

chiarazampolli:
Hello @rihanphys ,

The error does not depend on your changes. I pinged already @ktf  and @MichaelLettrich on a PR with the same issue.

Chiara

ktf:
@ihrivnac I suspect your changes for geant4 broke the gun test. Could you have a look?

ihrivnac:
The errors are reported also for ROOT libraries:

```
calling rungen.sh
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libRIO, output: $ROOT_DYN_PATH/libRIO
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libCling, output: $ROOT_DYN_PATH/libCling
Error in <TSystem::ExpandFileName>: input: $ROOT_DYN_PATH/libSTEER, output: $ROOT_DYN_PATH/libSTEER

```
But let's try to revert my last update and see if it helps.

PR #1224, opened 2020-09-17
Particle type added for special reconstruction for triton and He3

mcolocci:


STEER/ESD/AliESDpid.cxx
```diff
@@ -36,6 +36,7 @@
 #include "AliESDtrack.h"
 #include "AliMCEvent.h"
 #include "AliTOFPIDParams.h"
+#include "AliPID.h"
```
chiarazampolli, line 39: Is this include needed? It was not there before, and AliPID was used.

chiarazampolli:
Apart from a doubt on an include (which is in case harmless), to me it looks fine. @shahor02 , since you followed the development by @mpuccio some time ago, is it fine for you to merge?

shahor02 (approved):

STEER/ESD/AliESDpid.cxx
shahor02: If the include was coming from other dependencies, its explicit inclusion will not do any harm, so, it is ok

shahor02:
+1

chiarazampolli:
Yes, I also wrote that it is anyway harmless (but I asked Manuel, for completeness). What I wanted you to check was the rest :-) 
Thanks for taking a look!

Chiara

STEER/ESD/AliESDpid.cxx
chiarazampolli: Yes, sure, I would have avoided it if not needed, but I totally agree. I was curious to know why @mcolocci added it.

shahor02:
@chiarazampolli I've looked the code also, looks fine. The class version does not need to be incremented if only static data member is added, but it also will not harm.

chiarazampolli:
Thanks Ruben!

Error expected and unrelated.

Chiara

STEER/ESD/AliESDpid.cxx
mcolocci: Ciao @chiarazampolli. About the AliPID dependence could be indeed avoided. I added at the beginning because we were seeing an error during the rebuild and at the first glance we thought they were related to it. Then, when I managed to rebuild fixing the error who was somewhere else…, I simply forgot to remove it. Thanks for checking!

chiarazampolli:
Hello @ktf , 

Will this PR go through the tests automatically after the change by @ihrivnac in alidist (https://github.com/alisw/alidist/pull/2488)?

Cheers,

Chiara

PR #1226, opened 2020-09-20
Option to swicth off Ds->K*K decays

fprino:
This option is added to increase the statistics of Ds->phi+pi decays in MC productions, see ALIROOT-8502


shahor02:
+1

PR #1242, opened 2020-09-24
Corrects lack of median background subtraction on the GA Patch

moliver813:
Corrects lack of median background subtraction on the GA Patch. The factor of 64 should be correct because the factor ends being (2*2) / (8*8), and the GA patches are 2x2 FastORs and the background patches are 8x8 FastORs.


mfasDa:
+1

Thanks @moliver813 for the fix!

PR #1245, opened 2020-10-01
Update nuclei masses with the latest CODATA release

mpuccio:
Using 2018 values summarised in https://physics.nist.gov/cuu/pdf/wall_2018.pdf

This is particularly important for Tritons and Helium-3, that were off my a measurable amount


chiarazampolli:
+1

chiarazampolli:
I did not check that the numbers are correct, I rely on @mpuccio .

PR #1246, opened 2020-10-02
Propagate number of MC events to generate to alien plugin in case of …

mfasDa:
…MC loop

Propagating the number of MC events allows that
add macros for generators can make use of this
information. This is particularly of relevance for
generators which are running as external processes
not steered by the event loop. Adding this information
does not impose any problem on existing
implementations


mfasDa:
How can this be that PR from me in this directory are automatically merged?

mfasDa:
Error unrelated

PR #1248, opened 2020-10-02
Added number of MC events in analysis manager.

mgheata:
This PR adds a new persistent data member to the analysis manager, accessible via: GetNMCevents(). This is filled in the lego train by the grid handler, and made available in both testing and grid running.


chiarazampolli:
+1

mfasDa:
I just tested with my testsetup and the number of MC events is properly propagated to the worker so I can access it in the add macro for the generator, so the fix seems to work.

pzhristov:
+1

PR #1251, opened 2020-10-02
fix compilation warning on Root6, conflict between /* and // comments…

gconesab:
… applied at the same time


gconesab:
+1

gconesab:
Hi,
this is a harmless commit, can this be forced? Thanks

PR #1252, opened 2020-10-07
Fixing case when specific version is required to find an entry.

chiarazampolli:


chiarazampolli:
+1

chiarazampolli:
Hello @pzhristov ,
Would you agree to merge this PR even if macos is not happy? As we discussed, the failure is unrelated to the PR itself.
Cheers,
Chiara

PR #1262, opened 2020-10-15
FT0 back;  it was removed accidentally

AllaMaevskaya:


AllaMaevskaya:
But it not related to my code

Atlantic777:
FYI, the XRootD problem appeared because of this https://github.com/alisw/alidist/pull/2545

PR #1267, opened 2020-10-20
User trigger implementation for PYTHIA8

lucamicheletti93:
User trigger added to AliGenPythiaPlus.cxx and fix of the function streaming issue


chiarazampolli:
+1

PR #1286, opened 2020-11-25
fixed overlap

AllaMaevskaya:
Fix overlap between FT0 radiators and FV0 support


AllaMaevskaya:
Can one tell me how I can check what is wrong. When I press detail button  it says "No errors found" and no logs available
Best regards
Alla

PR #1299, opened 2020-12-03
update Run3 equivalent cuts - golden chi2 is now replayed also in O2

mkruegerGitHub:


mkruegerGitHub:
@jgrosseo this I forgot to update after we added the legacy filter bit for the goldenChi2 to the Run3 converter

jgrosseo:
+1

PR #1300, opened 2020-12-04
Additional protection, needed for FLUKA

fprino:


pzhristov:
+1

ktf:
@TimoWilken AliRoot/release seems stuck...

TimoWilken:
@ktf `git ls-remote --heads https://github.com/google/benchmark` got stuck, it's been doing nothing since 06:38 this morning. I've killed it, it's now working again.

That stuck git command was running under the main aliBuild command, which has a 10h timeout -- which is why it wasn't killed earlier.

ktf:
Ok. We should add a per command timeout to all those in alibuild.

PR #1303, opened 2020-12-16
Safety check for monopoles with charge 0

miweberSMI:


miweberSMI:
Hi @miranov25,

this special safety check is needed for monopoles (not sure why not spotted before). 

Thanks and cheers,
Michael

miranov25:
+1
Hello Michael.

In case it is working in your test, than OK. 
Looks like it will not affect anything else but monopoles.

Marian

miweberSMI:
Hi @miranov25 

yes that is correct, it only affects monopoles (which only we are simulating) and test was fine with it.

Thanks and cheers,
Michael

miranov25:
Hello @miweberSMI 
I checked the souce code. In case floating exception is switched OFF (root.exe instead of AliRoot)
the code was working and providing correct results.
BB was not used for monopoles
I prefered to do not calculate BB for monopoles at all  instead of putting there dummy mass

miweberSMI:
Hi @miranov25,
I am not sure I understand your message, how can we make full simulations work without using aliroot?
Maybe you can let me know what you exactly suggest?
Thanks and cheers,
Michael

miranov25:
Hello @miweberSMI 

I agree, I was not clear. 
### Floating point exception in AliRoot/Root:
In case we use AliRoot libraries running Root, floating point exception is not fatal. But in AliRoot it is.
You can try to run in root/aliroot prompt:
```1/0.```
In root you will get inf. In aliroot inf and floating point exception.
I assumed that previous simulation was done with root, or floating point exception was switched off using other method.

At the end for the digitization nel was used anyway:
https://github.com/alisw/AliRoot/blob/2153a00a6c938fd63f386a32012adc578ed81050/TPC/TPCsim/AliTPCv2.cxx#L2549

What i suggested was to do not calculate dEdx at all.
Lines between:
* https://github.com/alisw/AliRoot/blob/2153a00a6c938fd63f386a32012adc578ed81050/TPC/TPCsim/AliTPCv2.cxx#L2530
* https://github.com/alisw/AliRoot/blob/2153a00a6c938fd63f386a32012adc578ed81050/TPC/TPCsim/AliTPCv2.cxx#L2547

are not relevant for monopoles as they used other dEdx model.

They should be in else part of block:
```
   if(pdg==kPdgMonopole){
 nel = (Int_t)((mc->Edep()-Emin)/wIon) + 1;
    }
    else{
       // caclualte standard stuff
```

miweberSMI:
Hi @miranov25, 

thanks for your further comments. I removed now the dummy charge and moved all non-relevant variable calculations into the "else" part. I checked with a monopole and pion gun and it seems to be OK. Can you give it another look?

Thanks and cheers,
Michael

miranov25:
Hello @miweberSMI 

New version looks better. In diff and source code I do not see error, but I suggest to
check not only monopoles but also some standard simulation.
Results should be exactly the same as in version before  (in case of the same random seed).

I had before my standard test, you can use something from the AliDPG

Regards
Marian

miweberSMI:
Hi @miranov25,

I was running two full DPG simulations with Pythia6 generator and 10 events locally using the same seed. Once with the master and once with inclusion of this PR. The timing of both is the same (+-1 sec) and a few selected TPC observables (plotted from the AODs) are also the same. So as far as I can tell, the PR is not affecting the simulation of non-monopole particles. If you have any specific suggestion, what to check in addition, please let me know.

Thanks and cheers,
Michael

miranov25:
+1

PR #1306, opened 2021-01-20
Move TPCSpaceChargeBase from GPU to TPC

davidrohr:
This has been removed from O2/GPU, so have to create a copy in AliRoot before we sync the GPU folder


davidrohr:
+1

PR #1312, opened 2021-01-28
ATO-520 - BUG fixex + make default setting working

miranov25:
* using reasonable for fit ranges
* set initial range of the fit fuction to histogram range
* adding NDF of the  to the output stream
* Bug fix - Normalization of the chi2 to NDF


miranov25:
+1

PR #1313, opened 2021-02-02
[doxygen] Add documentation and fix formatting

mfasDa:
- Add documentation about the definition of the
  median array - this is crucial for user to understand
  which median value to use in the background
  subtraction for either of the two subdetectors
- Fix indentation to 2 spaces


mfasDa:
@pzhristov Could you please merge the commit even though this is in STEER? This is a trivial commit, only formatting and documenting, no change of code. Specifications of parameters must be well documented in order to make the code useful.

chiarazampolli:
+1

PR #1315, opened 2021-02-04
EMCal Trigger Digit overwrite fix

moliver813:
Fixes an error where new L0 patch digits would be created with the wrong fastOR id.

This led to an apparent trigger inefficiency when multiple digits in the same location overwrote each other.


mfasDa:
+1

PR #1317, opened 2021-02-12
Modifications to initialize TF1 from string and TVectorD

wiechula:
To allow for forward and backward compatibility between root 5 and root 6, the sigma parametrization function is initialized now from a string for the function definition and a TVectorD for the fit parameters.
Function and parameters are set via an input file to the
MakeTPCPIDResponseOADB macro.


chiarazampolli:
+1

miranov25:
As we discussed today with Jens, I will add and check qPt:tgl correction which I developed in December.
In case the QA/dashboard  test will be successfull i will make next pull request

PR #1321, opened 2021-02-18
Adds noise to EMCal L1 Trigger set via SimParams

moliver813:
Updates SimParams, adding L1ADCNoise param
Adds noise to L1 trigger timesums based on param in SimParam OCDB

### DIFF
--- a/EMCAL/EMCALbase/AliEMCALSimParam.cxx
+++ b/EMCAL/EMCALbase/AliEMCALSimParam.cxx
@@ -19,6 +19,8 @@
 #include "AliEMCALSimParam.h"
 #include "AliLog.h"
 
+#include <iostream>
+
 /// \cond CLASSIMP
 ClassImp(AliEMCALSimParam);
 /// \endcond
@@ -42,7 +44,8 @@ fTimeResolutionPar1(0),
 fNADCEC(0),//Digitizer
 fA(0.),
 fB(0.),
-fECPrimThreshold(0.) //SDigitizer   
+fECPrimThreshold(0.), //SDigitizer
+fL1ADCNoise(0.0)//TriggerElectronics
 {
   // Parameters in Digitizer
   fMeanPhotonElectron = 4400;    // electrons per GeV 
@@ -61,12 +64,13 @@ fECPrimThreshold(0.) //SDigitizer
   fA                  = 0;
   fB                  = 1.e+6; // Dynamic range now 2 TeV
   fECPrimThreshold    = 0.05;  // GeV	// threshold for deposit energy of hit
+
 }
 
 ///
 /// Copy constructor.
 //-----------------------------------------------------------------------------
-AliEMCALSimParam::AliEMCALSimParam(const AliEMCALSimParam& ):
+/*AliEMCALSimParam::AliEMCALSimParam(const AliEMCALSimParam& ):
 TNamed(),
 fDigitThreshold(0),
 fMeanPhotonElectron(0),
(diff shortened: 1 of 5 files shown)


gconesab:
Hi, I leave this to Markus, but why have you removed the copy constructors? I am not sure they are ever needed but they do not harm.

mfasDa:
@gconesab Singleton classes must never have a copy constructor - they violate the idea behind a singleton that there can never be two instances at the same time.

EMCAL/EMCALsim/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ include_directories(${AliRoot_SOURCE_DIR}/EMCAL/${MODULE})
 # Additional include folders in alphabetical order except ROOT
 include_directories(SYSTEM ${ROOT_INCLUDE_DIR})
 include_directories(${AliRoot_SOURCE_DIR}/EMCAL/EMCALbase
+                    ${AliRoot_SOURCE_DIR}/EMCAL/EMCALTriggerBase
```
mfasDa, line 26: Eventually we need to move the constants file to EMCALutils, not sure about the dependency here. I would prefer to have the dependencies as lightweight as possible.

EMCAL/EMCALbase/AliEMCALSimParam.cxx
```diff
@@ -131,5 +134,9 @@ void AliEMCALSimParam::Print(Option_t *) const
   printf("\t sdigitization parameters       A = %f\n",     fA);
   printf("\t                                B = %f\n",     fB);
   printf("\t Threshold for EC Primary assignment  = %f\n", fECPrimThreshold);
+
+  printf("=== Parameters in TriggerElectronics === \n");
+  printf("\t L1 FastOR timesum noise (fL1ADCNoise) = %f\n", fL1ADCNoise);
+
```
mfasDa, line 140: Please rather implement operator<< for AliEMCALSimParams - we are in the c++ time and the time for such print functions are over. We usually do the following:

- Add PrintStream functions (in header: include iosfwd instead of iostream) with the following interface
``` c++
void PrintStream(std::ostream &stream) const;
```
  and implement the printing in the same way as if one would use std::cout.
- Implement operator<< becomes now trivial:
``` c++
std::ostream &operator<<(std::ostream &stream, const AliEMCALSimParam &param) {   
  param.PrintStream(stream);
  return stream;
}
```
- Finally the print function becomes:
``` c++ 
void AliEMCALSimParam::Print(Option_t *) {
  std::cout << *this << std::endl;
}
```

This adds more flexibility to the debug streaming as besides the print function this can be used in std::cout, but better with any kind of logging library like the loggers in AliLog (AliErrorStream, AliWarningStream, AliInfoStream, AliDebugStream) ... where log levels can be set

mfasDa (changes requested):
Please have a look at the logging, and eventually we need to fix the dependency chain here.

EMCAL/EMCALbase/AliEMCALSimParam.cxx
moliver813: For the Print(Option_t *) function, should it be AliInfoStream() or AliDebugStream() instead of std::cout?

mfasDa: This is basically a compatibility function to ROOT5 style logging where I think std::cout is the way to go.

mfasDa:
+1

ktf:
+1

I think it's because a CMakeLists is modified, but I am not 100% sure.

mfasDa:
@ktf But these CMankeLists within EMCAL, so we cannot even update our own CMakeLists any more?

moliver813:
I just added #include <iostream> to the AliEMCALSimParams.h. I think the absence of that was causing the build issue in the MacOS build.

mfasDa:
+1

mfasDa:
The approval should be withdrawn for the moment, as AliEMCALSimParams.h now includes both iostream and iosfwd, however it should only include iosfwd and must under no circumstance include iostream. Waiting for a fix before the PR can be merged.

moliver813:
Just submitted a fix that moves the iostream include to AliEMCALSimParam.cxx

mfasDa:
+1 

Thanks @moliver813 for fixing! Should be fine now.

@ktf Could you please approve once more?

ktf:
+1

PR #1322, opened 2021-02-19
Allow for running tests outside ALICE environment

mfasDa:
Currently running ctest stand-alone requires initializing
the ALICE environment outside. Explicitly specifying PATH
and ROOTSYS are sufficient to launch ctest without
being in an ALICE environment.


mfasDa:
Any problem with this? On the usual build test it doesn't hurt, but it can be quite useful.

chiarazampolli:
Hello @ktf , 

Do you agree to approve?

Chiara

mfasDa:
Would be good if at the same time also the corresponding pull request in AliPhysics could be processed: https://github.com/alisw/AliPhysics/pull/17042

mfasDa:
At least launching the test would not harm ...

mfasDa:
Let's see whether this will ever have a chance to be merged ...

The corresponding commit in AliPhysics is!

ktf:
+1

PR #1324, opened 2021-02-25
A side Cables

AllaMaevskaya:


AllaMaevskaya:
I am sorry but I don't understand reason of crash

chiarazampolli:
Hello @ktf , @TimoWilken , 

Any idea about the errors here?

Chiara

TimoWilken:
@chiarazampolli @ktf I don't know why `rootcint` segfaults when compiling some headers here, sorry! It crashes when compiling a header that was changed in this PR (it compiles all the `AliFITv*.h` headers; v8 was changed here), but I don't know enough about `rootcint` to tell whether this PR is causing the crash or whether it's unrelated.

AllaMaevskaya:
But headers did not changed in AliFITv8

TimoWilken:
@AllaMaevskaya sorry, that wasn't clear -- I meant that this PR changes the file `FIT/FITsim/AliFITv8.h` ([see changes here]), and that is one of the files that `rootcint` is compiling when it crashes. That makes me suspicious :)

I don't know enough about what `rootcint` does to know whether these changes cause the crash or not, though.

[see changes here]: https://github.com/alisw/AliRoot/pull/1324/files#diff-032d1e3bb61eaefbd3afed98c82618185d2e6280d577b2f59c923be2436f5a33L28

AllaMaevskaya:
I can not repeat this crash with UBUNTU 20.04 and ROOT6. And what is interesting  -  on macos there was not crash

chiarazampolli:
Hello,

Could it be some memory issue in the CI? 

Chiara

TimoWilken:
Hm, I'm not sure about that. Running out of memory doesn't usually cause segfaults; instead, offending process is just killed. This would give a very different message in the logs. Normally segfaults are caused by a compiler bug.

chiarazampolli:
Hello, 
Just looking in the code, it is not so easy for me to debug, since you changed several things, arrays etc (BTW, you did not follow the convention that data members should start with "f", see lines from 91 to 101 of https://github.com/alisw/AliRoot/blob/2f5a6a9d25966297d7afc724cdb50d06ba920dd0/FIT/FITsim/AliFITv8.h). 

@AllaMaevskaya , can you check that things are consistent? Even though if they were not, should not macos also complain?

Chiara

AllaMaevskaya:
I fixed coding convenience.
macos did not complain

chiarazampolli:
Hello Alla,

I did not mean that the coding conventions were the origin of the issue, but thanks for fixing them. You checked that all array sizes are fine? Even though I am not sure that this would cause the error we see. 
I am surprised too that the macos works.
@TimoWilken , @ktf , do you have any suggestion on how to check?

Chiara

chiarazampolli:
Hello,

After the latest change by Alla (who fixed some array definition), we still have the same error. Maybe @davidrohr , or @shahor02  could also know what is wrong. 

Chiara

TimoWilken:
Hm, I get exactly the same segfault in an slc7 container locally (with plenty of memory). I suspect this is a bug in rootcint...

chiarazampolli:
Hello @TimoWilken , @ktf ,

How should we proceed? If we merge trusting macos, then all following PRs in AliRoot will fail in release, right?

Chiara

ktf:
I would not "trust macOS". If @TimoWilken managed to reproduce it standalone on slc7, I would say it's something it needs to be debugged and fixed. Adding @pzhristov as well.

chiarazampolli:
I could not agree more, since Alla said that on Ubuntu 20 she does not see any problem, I don't know how to debug it.

FIT/FITsim/AliFITv8.cxx
```diff
@@ -2041,12 +2041,12 @@ void AliFITv8::SetCablesA(TGeoVolume *stl) {
   // 12 cables extending beyond the frame
   Float_t pcablesextend[3] = {2, 15, 0.245};
   Float_t pcablesextendsmall[3] = {3, 2, 0.245};
-  Float_t *ppcablesextend[] = {pcablesextend, pcablesextend, pcablesextendsmall,
+  Float_t *ppcablesextend[4] = {pcablesextend, pcablesextend, pcablesextendsmall,
```
ktf: Maybe you should try with `ppcablesextend[4][3]`?

chiarazampolli: Why not even just `ppcablesextend[12]` with a nice loop to fill the elements?

AllaMaevskaya:
Now I removed at all this cycle and it still crashes. I have no idea what can be done. I O2 the same code compiles and works

PR #1328, opened 2021-03-02
Add 5 charged and 5 neutral pentaquark states

JakeMartinez:
adding Ps states to run an MC to get acc x eff for yield upper limit corrections.

### DIFF
--- a/STEER/STEER/AliMC.cxx
+++ b/STEER/STEER/AliMC.cxx
@@ -395,8 +395,8 @@ void  AliMC::AddParticles()
   TVirtualMC::GetMC()->DefineParticle( 9322136, "Ps_2500", kPTHadron, 2.5 ,  1.0, 1.6455e-23,"Hadron", 4.e-2, 5, 1, 0, 0, 0, 0, 0,  1, kTRUE);
   TVirtualMC::GetMC()->DefineParticle(-9322136, "AntiPs_2500", kPTHadron, 2.5 , -1.0, 1.6455e-23,"Hadron", 4.e-2, 5, 1, 0, 0, 0, 0, 0, -1, kTRUE);
 
-  Int_t psmode[10][3] = {0};
-  Float_t psratio[10] = {0.f};
+  Int_t psmode[7][3] = {0};
+  Float_t psratio[7] = {0.f};
   psratio[0] = 100.;
 
   psmode[0][0] = 333;  // phi
@@ -410,7 +410,7 @@ void  AliMC::AddParticles()
 
 	
   //Ps+- and Ps0 Hidden Strange Pentaquarks
-  for (int j=0; j<10; j++) {
+  for (int j=0; j<7; j++) {
   psmode[j][0] = psmode[j][1] = 0;
   psratio[j] = 0.;
       }
@@ -519,7 +519,7 @@ void  AliMC::AddParticles()
     
     
   // Anti-Ps0
-  for (int j=0; j<10; j++) {
+  for (int j=0; j<7; j++) {
   psmode[j][0] = psmode[j][1] = 0;
   psratio[j] = 0.;
   }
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hello @JakeMartinez , 

Maybe this is a naive question, but if you add 5 states, why do you increase the vectors by 4, from 6 to 10? I admit that I don't know how this piece of code works exactly, so it might be fine.

Chiara

miweberSMI:
Hi @chiarazampolli,
I was wondering the same. The vectors are for the decays though. But @JakeMartinez added maximal 7 decay chains, so I think increasing it by one should be sufficient?
Cheers,
Michael

STEER/STEER/AliMC.cxx
```diff
@@ -408,8 +408,227 @@ void  AliMC::AddParticles()
...
+  psmode[3][0] = 3114; // Sigma*-
+  psmode[3][1] = 310; // K0
+  psratio[3] =  16.24045534;
+  psmode[4][0] = 2212; // Proton
+  psmode[4][1] = 333; // Phi
+  psratio[4] = 12.48972884;
+  psmode[5][0] = 3122; // Lambda
+  psmode[5][1] = 323; // K*+
+  psratio[5] = 49.90179770;
+  psmode[6][0] = 3124; // Lambda(1520)
```
miweberSMI, line 489: for example here

chiarazampolli:
I think he wants to add on top of what existed, but I am not sure... Let's see what @JakeMartinez answers.

JakeMartinez:
when I initially wrote this part of the code, I had charged and neutral states mixed together, and the maximum then was 10 channels. After separating them by charge (and adding 2 channels), now the maximum for any state is 7. I should have changed the vector size to 7 before submitting.

chiarazampolli:
Hello @JakeMartinez , 

Please, fix the PR, then we will take another look. 

Chiara

JakeMartinez:
I created another commit, changing the size from 10 to 7, hopefully this is the correct way to fix.

STEER/STEER/AliMC.cxx
```diff
@@ -408,8 +408,227 @@ void  AliMC::AddParticles()
...
+  psmode[2][0] = 3224; // Sigma*+
+  psmode[2][1] = 310; // K0
+  psratio[2] =  46.98510416;
+  psmode[3][0] = 3114; // Sigma*-
+  psmode[3][1] = 310; // K0
+  psratio[3] =  44.54185676;
+  TVirtualMC::GetMC()->SetDecayMode(9223132,psratio,psmode);
+    
+    
+  TVirtualMC::GetMC()->DefineParticle( 92231320, "Ps_0_1870", kPTNeutron, 1.87 ,  0.0, 6.582e-24, "Hadron", 0.1, 1, -1, 0, 1, -1, 0, 0,  1, kTRUE);
```
chiarazampolli, line 455: Before here, don't you need to re-initialize the psmode and psratio? At least you changed above the 3-rd element, which will then go also in the decay mode below.

chiarazampolli (commented):
Ciao @JakeMartinez, 

Thanks, the new commit fixed the problem, but I am afraid you need to re-initialize the vectors at some decay modes, see e.g. my comment below.

STEER/STEER/AliMC.cxx
JakeMartinez: Yes, I see what you are saying, good catch. Perhaps instead of reinitializing (which will work), I could rearrange from fewest to most decay channels? This will make for fewer lines and still passing wrong decay modes.

chiarazampolli: Hello @JakeMartinez,
I think that the cleanest is to reinitialize. 
Chiara

pzhristov:
+test

gconesab:
Hi all, can this be merged or are we still waiting for some extra fixes/developments?

PR #1340, opened 2021-03-15
Enable more than one file in chain for local analysis

pestratm:
When running a PCM analysis locally on a chain containing more than one AOD file, apparently no AOD Conversion Photons are extracted and, consequently, all histograms containing properties of the Conversion Photons remain empty. This error, however, does not occur when there is only one file in the chain. Introducing some modifications to the AliAODInputHandler::Init() and AliAODEvent::ReadFromTree() methods in AliRoot seems to fix this issue. In AliAODInputHandler::Init(), this is loading the first tree in the chain at the beginning. In AliAODEvent::ReadFromTree(), this is setting the branch addresses for every tree anew.

### DIFF
--- a/STEER/AOD/AliAODEvent.cxx
+++ b/STEER/AOD/AliAODEvent.cxx
@@ -733,17 +733,46 @@ void AliAODEvent::ReadFromTree(TTree *tree, Option_t* opt /*= ""*/)
     // Try to find AliAODEvent
   AliAODEvent *aodEvent = 0;
   aodEvent = (AliAODEvent*)tree->GetTree()->GetUserInfo()->FindObject("AliAODEvent");
+  
+  TList* friendL;
   if(aodEvent){
     // This event is connected to the tree by definition, just say so
     aodEvent->SetConnected();
       // Check if already connected to tree
     TList* connectedList = (TList*) (tree->GetUserInfo()->FindObject("AODObjectsConnectedToTree"));
     if (connectedList && (!strcmp(opt, "reconnect"))) {
+		
         // If connected use the connected list of objects
         if (fAODObjects != connectedList) {
            delete fAODObjects;
            fAODObjects = connectedList;
         }   
+        
+        friendL = tree->GetTree()->GetListOfFriends();
+		if (friendL) 
+		{
+		  // set the branch addresses
+		TIter next(fAODObjects);
+		TNamed *el;
+		while((el=(TNamed*)next())){
+		  TString bname(el->GetName());
+			// check if branch exists under this Name
+		  TBranch *br = tree->GetTree()->GetBranch(bname.Data());
+		  if(br){
+			tree->SetBranchAddress(bname.Data(),fAODObjects->GetObjectRef(el));
+		  } else {
+			br = tree->GetBranch(Form("%s.",bname.Data()));
+			if(br){
+			  tree->SetBranchAddress(Form("%s.",bname.Data()),fAODObjects->GetObjectRef(el));
+			}
+			else{
+			  printf("%s %d AliAODEvent::ReadFromTree() No Branch found with Name %s. \n",
+					 (char*)__FILE__,__LINE__,bname.Data());
+			}	
+		  }
+		}
+	}
+        
         GetStdContent(); 
         fConnected = kTRUE;
         return;
@@ -764,7 +793,7 @@ void AliAODEvent::ReadFromTree(TTree *tree, Option_t* opt /*= ""*/)
     }
       //
       // Let's find out whether we have friends
-    TList* friendL = tree->GetTree()->GetListOfFriends();
+    friendL = tree->GetTree()->GetListOfFriends();
     if (friendL) 
     {
       TIter next(friendL);
--- a/STEER/AOD/AliAODEvent.cxx
+++ b/STEER/AOD/AliAODEvent.cxx
@@ -741,7 +741,7 @@ void AliAODEvent::ReadFromTree(TTree *tree, Option_t* opt /*= ""*/)
       // Check if already connected to tree
     TList* connectedList = (TList*) (tree->GetUserInfo()->FindObject("AODObjectsConnectedToTree"));
     if (connectedList && (!strcmp(opt, "reconnect"))) {
-		
+
         // If connected use the connected list of objects
         if (fAODObjects != connectedList) {
            delete fAODObjects;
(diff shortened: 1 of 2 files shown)


pzhristov:
Hi @mgheata would you review and test these changes?

mgheata:
Hi, I will review the code. We need an example as test case, it is not clear for me how other analysis using AOD friends and chains was working before (e.g. VertexingHF analysis.

STEER/AOD/AliAODEvent.cxx
```diff
@@ -733,17 +733,46 @@ void AliAODEvent::ReadFromTree(TTree *tree, Option_t* opt /*= ""*/)
...
         if (fAODObjects != connectedList) {
            delete fAODObjects;
            fAODObjects = connectedList;
         }   
+        
+        friendL = tree->GetTree()->GetListOfFriends();
+		if (friendL) 
+		{
+		  // set the branch addresses
+		TIter next(fAODObjects);
```
mgheata, line 755: Looks like the indentation is wrong, may be due to use of tabs instead of spaces, can you fix it?

STEER/AOD/AliAODEvent.cxx
```diff
@@ -733,17 +733,46 @@ void AliAODEvent::ReadFromTree(TTree *tree, Option_t* opt /*= ""*/)
...
+			br = tree->GetBranch(Form("%s.",bname.Data()));
+			if(br){
+			  tree->SetBranchAddress(Form("%s.",bname.Data()),fAODObjects->GetObjectRef(el));
+			}
+			else{
+			  printf("%s %d AliAODEvent::ReadFromTree() No Branch found with Name %s. \n",
+					 (char*)__FILE__,__LINE__,bname.Data());
+			}	
+		  }
+		}
```
mgheata, line 773: It seems to me that this block was executed before in all cases (a bit below), so why adding it again will change something?

STEER/AOD/AliAODInputHandler.cxx
```diff
@@ -106,7 +106,8 @@ Bool_t AliAODInputHandler::Init(TTree* tree, Option_t* opt)
 
     fTree = tree;
     if (!fTree) return kFALSE;
-    GetEntries();
+    //GetEntries();
```
mgheata, line 109: Why having to comment this out? It is currently used to check the friend open status stored in the chain by TChain::GetEntries(). Disabling it will make grid jobs with bad files crash only when reaching those, instead of crashing in the beginning of the job.

mgheata (changes requested):

STEER/AOD/AliAODEvent.cxx
pestratm: This should be fixed now

pestratm:
Short illustration on what I have tried to do: I tried to locally run a PCM analysis (using the AliAnalysisTaskGammaConvV1 class 
of AliPhysics). For this, I create a chain of (several) AOD files using CreateAODChain.C that I passed to AliAnalysisManager

     mgr->StartAnalysis("local", chain); If the AOD chain contains only one file, this works without any problems. However, if the chain contains more than one file, no conversion photons are extracted and corresponding histograms remain empty or show zero conversion candidates for any event

STEER/AOD/AliAODInputHandler.cxx
pestratm: In TChain::GetEntries(), the last tree gets loaded `const_cast<TChain*>(this)->LoadTree(TTree::kMaxEntries-1);`. Shouldn't in the first step the first tree be loaded (thus the LoadTree(0) I added)?

STEER/AOD/AliAODEvent.cxx
pestratm: After loading the first tree in AliAODInputHandler::Init, now instead of not extracting any photon conversion candidates, it now extracted them only for the first tree. So it seemed to me that only the branch addresses of the first tree are set properly. Setting them again for every subsequent tree fixed this issue for me.

STEER/AOD/AliAODInputHandler.cxx
mgheata: No, LoadTree(TTree::kMaxEntries-1) forces the chain to discover the number of entries in every file from the chain. This is needed for the chain to setup its internals, and now we also use the status stored per file entry in the chain to detect if the file could be opened or not. If we comment out GetEntries, this is not checked and the job will crash during the event loop (I tested this). Normally LoadTree() gets called automatically by GetEntry() (called by the selector in the event loop), so I don't see why adding LoadTree(0) is needed

mgheata:
> Short illustration on what I have tried to do: I tried to locally run a PCM analysis (using the AliAnalysisTaskGammaConvV1 class
> of AliPhysics). For this, I create a chain of (several) AOD files using CreateAODChain.C that I passed to AliAnalysisManager
> 
> ```
>  mgr->StartAnalysis("local", chain); 
> ```
> 
> If the AOD chain contains only one file, this works without any problems. However, if the chain contains more than one file, no conversion photons are extracted and corresponding histograms remain empty or show zero conversion candidates for any event

Sure, but is this analysis available somewhere, or do you have a test macro that you can share?

STEER/AOD/AliAODEvent.cxx
mgheata: Maybe, but I see that the lines that you added  are the same as the ones below at [lines 823-838]( https://github.com/alisw/AliRoot/blob/d755faa53c655315b4e58f6eb0ba61524eff4987/STEER/AOD/AliAODEvent.cxx#L823-L838) so this code was called anyway, now it is called twice, or am I missing something?

pestratm:
> > Short illustration on what I have tried to do: I tried to locally run a PCM analysis (using the AliAnalysisTaskGammaConvV1 class
> > of AliPhysics). For this, I create a chain of (several) AOD files using CreateAODChain.C that I passed to AliAnalysisManager
> > ```
> >  mgr->StartAnalysis("local", chain); 
> > ```
> > 
> > 
> > If the AOD chain contains only one file, this works without any problems. However, if the chain contains more than one file, no conversion photons are extracted and corresponding histograms remain empty or show zero conversion candidates for any event
> 
> Sure, but is this analysis available somewhere, or do you have a test macro that you can share?

I attached the macro that I have been using (startANAROOT6.sh). In this case, I used data from run# 286350, pass1 of LHC18d, but the same problem occurred for any other data I used, too. 
[LocalTrainTests.zip](https://github.com/alisw/AliRoot/files/6295501/LocalTrainTests.zip)

STEER/AOD/AliAODEvent.cxx
pestratm: It is indeed the same code. But if I see correctly, that part is only executed for the first tree because of the `if (connectedList && (!strcmp(opt, "reconnect")))` statement. So for any subsequent tree, the connected list exists but the branch addresses are not set correctly, or are they?

mgheata:
> I attached the macro that I have been using (startANAROOT6.sh). In this case, I used data from run# 286350, pass1 of LHC18d, but the same problem occurred for any other data I used, too.
> [LocalTrainTests.zip](https://github.com/alisw/AliRoot/files/6295501/LocalTrainTests.zip)

I could run your test and noticed that in your macro the line `chain = CreateAODChain(localFiles.Data(), numLocalFiles,0,kFALSE,"AliAODGammaConversion.root");` was commented, and the chain was created without friends. You are setting `aodH->AddFriend` when you create the handler, but this is not sufficient since `AliAODHandler::ConnectFriends` seems to work only for the single tree mode - this is why you get correct results for a single tree. `SetBranchAddress` for a chain does not need to be called individually for each tree inside, the method is called only for the chain, which takes care internally when switching the current tree - so your reconnection of branches is redundant. So I'm not convinced that the patch is really needed (or at least in this form), maybe we should understand better the problem first, since it affects all the analysis.

pestratm:
Even when i use `chain = CreateAODChain(localFiles.Data(), numLocalFiles,0,kFALSE,"AliAODGammaConversion.root");` instead of `chain = CreateAODChain(localFiles.Data(), numLocalFiles);`, I still have the same problem. 
What I noticed was that, initially, I didn't get any photon conversion candidates for multiple trees. After adding `LoadTree(0)` in `AliAODInputHandler::ConnectFriends`, I got the photon conversion candidates of the first tree in the chain. After reconnecting the branches in `AliAODEvent::ReadFromTree`, I eventually got all photon conversion candidates from every tree. Could this be because when `AliAODEvent::ReadFromTree` is called,  the friend trees aren't properly connected to all trees yet so it can't set the branches for all trees yet? 
I also noticed that with the original code, for multiple trees in the chain, I got the message 
`<-AliAnalysisSelector->Init()`
`E-TTree::SetBranchStatus: unknown branch -> GammaConv_00000003_06000008d00100001100000000_gamma`
`E-TTree::SetBranchStatus: unknown branch -> GammaConv_00000003_16000008d00100001100000000_gamma`
`->AliAnalysisSelector::Process()`
which did not appear if I just had a single tree in the chain.

mgheata:
For the friends of the chain to work properly, one needs to call `chainFriend->Add(friendfile)` for each `chain->Add(file)` at chain creation time (see an example of how `CreateChain` is generated by the plugin in the trains [here](http://alitrain.cern.ch/train-workdir/PWGGA/GA_pp_AOD/1773_20210412-1602/__TRAIN__/lego_train.C)). What `AliAODInputHandler::ConnectFriends` does is just calling `AddFriend` for the current tree in the chain (which is not even the first one in case of multiple files due to the call to `chain->GetEntries`). So you should add systematically the friend chain at creation time. Why it still does not work downstream in your code even when doing this I don't know, but it clearly works in many trains that we can check in the lego system. Just speculating, but could it be that you read some non-standard branches (that are not listed in AliAODEvent::fAODObjects)? If you can prepare maybe a simpler test making just the chain and reading one branch needed in this analysis, it would be easier to debug.

PR #1341, opened 2021-03-21
full A-side material

AllaMaevskaya:


AllaMaevskaya:
error message does not match source code. 
FIT/FITsim/AliFITv8.h:488 has no sXoffset as seen on https://github.com/alisw/AliRoot/blob/7c78089432d4ec7354bd76b24002be3206123026/FIT/FITsim/AliFITv8.h#L488

I can not reproduce crash locally with Ubuntu 20.04 and o2-default

chiarazampolli:
Ciao Alla,

Could it be that they need to be defined as doubles? See: https://github.com/alisw/AliRoot/blob/7c78089432d4ec7354bd76b24002be3206123026/FIT/FITsim/AliFITv8.h#L472

Chiara

chiarazampolli:
Ciao @AllaMaevskaya , 

So, this is finally merged. What was the issue? And why was it not seen on Ubuntu? I am asking to avoid the same in the future. 

Chiara

PR #1343, opened 2021-03-30
Fix: add protection against too small likelihoods container

shahor02:
@chiarazampolli this is a brute-force protection: depending on the options, the `countSpecies` can be as small as 3, while the https://github.com/shahor02/AliRoot/blob/9d935c6602778cdbf833c95d00eea23f50f47cc2/ANALYSIS/ANALYSISalice/AliAnalysisTaskPIDqa.cxx#L1514 is called with kSPECIES=5, overflowing the `likelihoods`.
This bug dates back to 2016, strange that it fired only now. The logics should be checked by the authors (Yvonne? did not find hew on the git)


chiarazampolli:
Hi @shahor02 ,

Thanks! I was just now fighting with gdb at the ComputeTRDprobability, wonderful that you found the issue before me. 
Maybe we see the issue since we run with Ubuntu 20? I don't know. 
I will implement your patch, and confirm that it works. 

Chiara

chiarazampolli:
Hi @shahor02 ,

Wonderful, I confirm that this fixes the crash (which was seen only when compiling in DEBUG mode, which is what I usually have).

Thanks!

Chiara

PR #1348, opened 2021-04-21
Fix application of TPC pileup correction for MC

jezwilkinson:
Add correct use of "fUseTPCPileupCorrection" flag to TPCPIDResponse::GetExpectedSignal call for Monte Carlo when running with TuneOnData.
Removes shift in mean of TPC distribution observed in LHC18qr MC with new splines.


chiarazampolli:
+1

PR #1350, opened 2021-04-28
Replace fatal with error for files with status -3 that cannot be open…

mgheata:
…ed, to allow the chain to skip bad files.


mgheata:
@jgrosseo can you have a look?

jgrosseo:
+1

PR #1351, opened 2021-04-28
Fixes from S. Aiola to the HepMCParser ALIROOT-8653

gconesab:


chiarazampolli:
+test

gconesab:
Hi  @maireiphc and @qgp,

can you have a look and if you agree test/approve?

Thanks ... Gustavo

EVGEN/AliGenExtFile.cxx
```diff
@@ -336,6 +337,90 @@ void AliGenExtFile::Generate()
...
+        << " and PDG = " << parent1->GetPdgCode() << std::endl;
+      }
+      else if (parent2 && !parent1) {
+        // Only the second mother exists
+        AliDebugStream(3) << "Only second mother exists. Using mother 2 with index " << iparent2
+        << " and PDG = " << parent2->GetPdgCode() << std::endl;
+        particle->SetFirstMother(iparent2);
+      }
+      else if (parent1 && parent2) {
+        // Both mothers exist. Select the mother with the larger PDG code
```
maireiphc, line 381: In case of 2 parents, I am wondering whether the choice to retain the mother with the highest PDG code could create some side effects or is immune of troubles...
Clearly one decision has to be made on what to do and this one is likely the least worse, but still.

Looking at the MC scheme of PDG (https://pdg.lbl.gov/2018/reviews/rpp2018-rev-monte-carlo-numbering.pdf), I am trying to hunt for troubles.
I do not find a good counter-example but I can neither grant to have a clear view nor being exhaustive.
My main field of concerns would be about : 
- rescatterings in the hadronic phase in AA (2->2, 3->2, etc)
- diquarks Vs baryons (both ID families O(10^3) )
- leptons+gauge bosons vs hadrons ID (gamma gamma in UPC, gamma A in p-Pb, ...)

I will not hold the merging, and going for a concrete test would be already good to move forward.
Just wanted to discuss here, if the problems are expected to be cured once and for all of if we should expect some later unexpected behaviour hitting us without notice.

gconesab: Hi Antonin, I agree with your concerns, the idea is we have this in master now and in a special tag to test for productions so that we can see how this changes the herwing case and also the EPOS comparing with already existing EPOS.

gconesab: I think you can do "test"

maireiphc:
+test

maireiphc (approved):

gconesab:
Hi @ktf , the test fails but I think it is not the cause, can you have a look?
Thanks

EVGEN/AliGenExtFile.cxx
```diff
@@ -112,9 +112,9 @@ void AliGenExtFile::Generate()
   //
   Double_t origin[3] = {0,0,0};
   Double_t time = 0.;
-  Double_t p[4];
+  Double_t p[4] = {0,0,0,0};
```
cholmcc, line 115: OK, but why not 

    Double_t p[] = {0.,0.,0.,0.};

?

EVGEN/AliGenExtFile.cxx
cholmcc: Why are we not keeping both mothers?  If a model produces particles through interactions (whatever that may be), then the event record presented to GEANT (or similar) should have that reflected.   Simply removing a mother gives the false impression that the daughter particle is from a decay - which it probably isn't and must not be interpreted that way.  

Nor do I see the point of removing a mother from the tree.   What do we gain from that?  It seems to me that we also run the risk of having mothers that are not attached to an "end-vertex" and are thus considered "final state" and propagated through the experiment.  

I really don't see the point of this at all.  Please enlighten me as to what we hope to achieve with this.

TEvtGen/THepMCParser/THepMCParser.cxx
```diff
@@ -436,154 +439,445 @@ string THepMCParser::ParseGenEvent2TCloneArray(HepMC::GenEvent * genEvent, TClon
...
+    if (switchBeamPart) {
+      //AliDebugGeneralStream("THepMCParser", 3) << "Switching to the tree of the other beam particle to look for mother particle." << std::endl;
+      continue;
+    }
+  } while (true);
+
+//  AliDebugGeneralStream("THepMCParser", 3) << "Particles in event: " << nParticlesInHepMC
+//  << ", added: " << arrayID
+//  << ", entries: " << array->GetEntries() << "/" << array->GetEntriesFast()
+//  << std::endl;
```
cholmcc, line 879: TL,DR 

It seems to me the strategy should be 

- Given a vertex 
  - Register all out-going particle 
    - Check in mapping from GenParticle* to array index
      - If mapping exist, get particle 
      - otherwise create new particle in clones array from GenParticle and set mapping
   - Set first and last daughter index on incoming particles via mapping from GenParticle* to index 
    - For each outgoing particle - check if they have an end vertex 
      - If they do, do this procedure on that vertex 

This will then recurs the tree (yes, recursive calls). 

What is "special" in this case are particles without a production vertex (presumably beam particles).

EVGEN/AliGenExtFile.cxx
gconesab: Agreed, I just followed the same way of coding of the previous lines, I can remove it.

EVGEN/AliGenExtFile.cxx
gconesab: Hi Christian, thanks for having a look,
as discussed in https://alice.its.cern.ch/jira/browse/ALIROOT-8653, there is a problem in a recent Herwig production. Since Herwig was used in the past in fast simulations, it was suggested to implement the private modifications that worked then, that are in this commit https://github.com/aiola/alice-fast-simulation/commit/18b78d05cfe7b1674acb673bec7bb8572322437f#diff-0625d73f5f714b7d1e8694970ff7b9a4 from Salvatore, that unfortunately is not anymore ALICE. I am not sure what is the reasoning behind either.
Gustavo

gconesab:
Hi @maireiphc, can you retest/approve?
Let's try to follow the original plan. Test these modifications in master and special tag and revert them if not useful or wrong.
Cheers ... Gustavo

gconesab:
Hi @ktf 

can you please have a look, I doubt the errors here are due to this commit.

Thanks ... Gustavo

EVGEN/AliGenExtFile.cxx
maireiphc: Hello Christian, Gustavo, (all)
 looking at Salvatore's code, I started with the idea that it is not possible to attach 2 parents or more to the vertex (AliRoot-side limitation ?  HEPmc-side limitation ?). Based on Christian's input, it seems that this is possible to to attach so, at least on paper.
@mfasDa or somebody from PWG-JE : do you have any memory on the justification for Salvatore's choice ?

mfasDa: Hi @maireihpc, (all),
I don't recall  what is the motivation for Salvatore's choice. I think the number of parent particles is an AliRoot simulation. In Herwig usually the hard scatterings have two mother particles, which are partons from the beam particles involved in the scattering - probably for SHERPA it is the same. I think starting from those the particles the rest of the chain has only one mother, which are the clusters (gluons) - pdg code 81 and 82, which can split and then fragment at some point.
Cheers
Markus

maireiphc (approved):

cholmcc:
Hi all, 

So I took the code I wrote above and made it work :-)  The strategy had to change a little bit, since mother particles are not guarantied to exist at the time one is dealing with a vertex.  

Note, the code below should work on both HepMC and HepMC3 (there's a bunch of SFINAE compat code in the beginning).  I tested with HepMC3 input because that's what I had on file.  

```c++
/*
 * For HepMC3:
 *
 *  g++ -DHEPMC3 hepmc2root.cc `root-config --cflags --libs` -lEG -lHepMC3 -o hepmc2root
 *
 * For HepMC 
 *
 *  g++ -DHEPMC2 hepmc2root.cc `root-config --cflags --libs` -lEG -lHepMC  -o hepmc2root
 */
#ifndef HEPMC2ROOT
#define HEPMC2ROOT
#include <TClonesArray.h>
#include <TParticle.h>
#include <TTree.h>
#include <TFile.h>
#include <map>
#include <set>
#include <vector>
#include <iostream>
#include <type_traits>

//====================================================================
//
// Below are hacks to make the code work the same for HepMC and HepMC3
//
/** 
 * Check if the class T has member function T::beam_particles (HepMC::GenEvent) 
 */
#if 1
template <typename T>
struct has_beams 
{
  typedef char yes[1];
  typedef char no[2];
  template <typename C> static yes& test(decltype(&C::beam_particles));
  template <typename C> static no&  test(...);
  enum { value = sizeof(test<T>(0)) != sizeof(yes) };
};
#else
template <typename T>
using has_beams=std::is_member_function_pointer<decltype(&T::beam_particles)>;
#endif

//--------------------------------------------------------------------
/** 
 * Get beam particles using SFINAE using above check for the member
 * function Event::beam_particles and Event has that member function. 
 */
template <typename Event,
	  typename std::enable_if<has_beams<Event>::value,void>::type* = nullptr>
auto get_beams(Event& event)
{
  return event.beams();
}

//--------------------------------------------------------------------
/** 
 * Get beam particles using SFINAE using above check for the member
 * function Event::beam_particles and Event does not have that member function. 
 */
template <typename Event,
	  typename std::enable_if<!has_beams<Event>::value,void>::type* = nullptr>
auto get_beams(Event& event)
{
  auto b = event.beam_particles();
  std::vector<typename decltype(b)::first_type> r;
  r.push_back(b.first);
  r.push_back(b.second);
  return r;
}

//____________________________________________________________________
/** 
 * Check to see if we have particles_in_const_begin member function 
 */
template <typename T>
struct has_in 
{
  typedef char yes[1];
  typedef char no[2];
  template <typename C> static no&   test(decltype(&std::remove_pointer<C>::type::particles_in_const_begin));
  template <typename C> static yes&  test(...);
  enum { value = sizeof(test<T>(0)) == sizeof(yes) };
};
//--------------------------------------------------------------------
/** 
 * Check to see if we have particles_out_const_begin member function 
 */
template <typename T>
struct has_out
{
  typedef char yes[1];
  typedef char no[2];
  template <typename C> static no& test(decltype(&std::remove_pointer<C>::type::particles_out_const_begin));
  template <typename C> static yes&  test(...);
  enum { value = sizeof(test<T>(0)) == sizeof(yes) };
};
//--------------------------------------------------------------------
template <typename Vertex, typename E=void> struct in_range;
template <typename Vertex>
struct in_range<Vertex,
		typename std::enable_if<has_in<Vertex>::value,void>::type> 
{
  in_range(Vertex& v) : _v(v) {}
  auto begin() const { return _v->particles_in().begin(); }
  auto end() const { return _v->particles_in().end(); }
  Vertex& _v;
};
template <typename Vertex>
struct in_range<Vertex,
		typename std::enable_if<!has_in<Vertex>::value,void>::type> 
{
  in_range(Vertex& v) : _v(v) {}
  auto begin() const { return _v->particles_in_const_begin(); }
  auto end() const { return _v->particles_in_const_end(); }
  Vertex& _v;
};

//--------------------------------------------------------------------
template <typename Vertex, typename E=void> struct out_range;
template <typename Vertex>
struct out_range<Vertex,
		 typename std::enable_if<has_out<Vertex>::value,void>::type> 
{
  out_range(Vertex& v) : _v(v) {}
  auto begin() const { return _v->particles_out().begin(); }
  auto end() const { return _v->particles_out().end(); }
  Vertex& _v;
};
template <typename Vertex>
struct out_range<Vertex,
		 typename std::enable_if<!has_out<Vertex>::value,void>::type> 
{
  out_range(Vertex& v) : _v(v) {}
  auto begin() const { return _v->particles_out_const_begin(); }
  auto end() const { return _v->particles_out_const_end(); }
  Vertex& _v;
};
//____________________________________________________________________
/** 
 * Check if the class T has member function T::particles_size (HepMC::GenEvent) 
 */
template <typename T>
struct has_size
{
  typedef char yes[1];
  typedef char no[2];
  template <typename C> static no&  test(decltype(&C::particles_size));
  template <typename C> static yes& test(...);
  enum { value = sizeof(test<T>(0)) == sizeof(yes) };
};

//--------------------------------------------------------------------
/** 
 * Get beam particles using SFINAE using above check for the member
 * function Event::beam_particles and Event has that member function. 
 */
template <typename Event,
	  typename std::enable_if<has_size<Event>::value,void>::type* = nullptr>
size_t get_size(Event& event)
{
  return event.particles().size();
}

//--------------------------------------------------------------------
/** 
 * Get beam particles using SFINAE using above check for the member
 * function Event::beam_particles and Event does not have that member function. 
 */
template <typename Event,
	  typename std::enable_if<!has_size<Event>::value,void>::type* = nullptr>
size_t get_size(Event& event)
{
  return event.particles_size();
}

  
//====================================================================
template <typename ParticlePtr,
	  typename VertexPtr,
	  typename FourVector,
	  typename Event>
struct GenEvent2Array
{
  using ParticleMap=std::map<ParticlePtr,Int_t>;
  using SeenVertex=std::map<VertexPtr,std::pair<Int_t,Int_t>>;

  ParticleMap fParticleMap;
  SeenVertex  fSeenVertex;

  /** 
   * Get TParticle index and pointer to TParticle object from our
   * mapping from HepMC3::GenParticle.  This returns a pair of -1 and
   * null if the particle isn't found.
   */
  std::pair<Int_t,TParticle*> Map(ParticlePtr p, TClonesArray* a) const
  {
    auto it = fParticleMap.find(p);
    if (it == fParticleMap.end())
      return std::pair<Int_t,TParticle*>(-1,0);
    
    Int_t      no = it->second;
    TParticle* pp = static_cast<TParticle*>((*a)[no]);

    return std::make_pair(no,pp);
  }
  /** 
   * Convert all HepMC3::GenParticle to TParticle objects stored in the TClonesArray. 
   * 
   * This will first generate the beam particles, and then next recurs
   * the tree to make all daughters.  Once that is done, we iterate
   * over the seen vertices and fill in the mother and daughter
   * information.  In this way, we are sure that
   *
   * - Daughters of the same vertex are consecutive in the array (as
   *   mandated by the HEPEVT format and adopted by ALICE).
   *
   * - Mothers a registerd.  Note, TParticle only has room for two
   *   mothers, and thus the first and last mother index fields of a
   *   TParticle indicated a _range_ of index between which all
   *   mothers can be found.  However, that does _not_ mean that all
   *   TParticle objects in that range _are_ mothers.  To verify that
   *   a particular mother candidate is indeed a mother, one must
   *   query the mother daughter range.  E.g., to find all mothers of
   *   a particle at some index, one would do
   *
   * @code 
   TList* AllMothers(Int_t id, TClonesArray* a)
   {  
     auto GetP = [a](Int_t i) { 
       return static_cast<TParticle*>((*a)[i]);
     };
     
     TParticle* p = GetP(id);
     
     TList* l = new TList;
     
     for (Int_t mid = p->GetFirstMother(); mid <= p->GetLastMother(); mid++) {
        TParticle* m = GetP(mid);
	if (m->GetFirstDaughter() >= id and m->GetLastDaughter() <= id) 
	  l.Append(m);
     }
 
     if (l.GetEntries() <= 0) {
       delete l;
       l = 0;
     }
   
     return l;
   }
   @endcode 
  */	
  bool Convert(Event& event, TClonesArray* array)
  {
    fParticleMap.clear();
    fSeenVertex.clear();
    array->Clear();

    auto beams = get_beams(event);
    
    // Check that we have beam particles at all 
    if (beams.size() <= 0) {
      std::cerr << "No beam particles for this event - cannot decode"
		<< std::endl;
      return false;
    }

    // First, we take all the beam particles
    for (auto p : beams)
      Convert(p, array);

    // Now, let's loop over all beam particles again and convert all
    // daughters connected to the end vertex.  This will traverse the
    // tree recursively.
    for (auto p : beams)
      Convert(p->end_vertex(), array);

    // Sanity check 
    if (array->GetEntries() != get_size(event))
      std::cerr << "Got " << get_size(event) << " but made "
		<<  array->GetEntries() << " particles" << " - something is off"
		<< std::endl;

    // Now loop over all vertices processed and fill in mother and daughter information. 
    for (auto v : fSeenVertex) {
      FillInDaughters(v.first, v.second, array);
      FillInMothers  (v.first, v.second, array);
    }
    
    return true;
  }

  /** 
   * Convert a single HepMC::GenParticle to TParticle and store 
   * mapping from that to index into TClonesArray 
   */
  Int_t Convert(ParticlePtr p, TClonesArray* a)
  {
    auto it = fParticleMap.find(p);
    if (it != fParticleMap.end())
      return it->second;
    
    VertexPtr  prodv = p->production_vertex();
    FourVector prod4;
    if (prodv) prod4 = prodv->position();
    FourVector mom4  = p->momentum();

    Int_t      no = a->GetEntries();
    TParticle* pp = new((*a)[no]) TParticle(p->pdg_id(), // Really should use pid
					    p->status(),
					    -1, -1, -1, -1,
					    mom4.px(),
					    mom4.py(),
					    mom4.pz(),
					    mom4.e(),
					    prod4.x(),
					    prod4.y(),
					    prod4.z(),
					    prod4.t());
    fParticleMap[p] = no;
    // std::cout << "Map " << p->id() << " -> " << no << std::endl;
    
    return no;
  }
  /** 
   * Generate all daughter particles of a vertex and recurs down the
   * tree from these dauhgters.  At the same time, we map the vertex
   * to first and last daughter index so we can set this on incoming
   * particles later.  Note, we have to do this in two steps, since
   * all mothers may not have been created yet.
   */ 
  bool Convert(VertexPtr v, TClonesArray* a)
  {
    // If no vertex, return 
    if (!v) return true;

    // If this vertex was already processed, return 
    if (fSeenVertex.find(v) != fSeenVertex.end()) return true;

    
    // Now we loop over the daughters
    Int_t first = 100'000'000;
    Int_t last  = -1;
    for (auto p : out_range<VertexPtr>(v)) { // ->particles_out()) {
      Int_t no = Convert(p,a);
      first = TMath::Min(first, no);
      last  = TMath::Max(last,  no);
    }
    fSeenVertex[v] = std::make_pair(first, last);
    
    if (first > last) {
      std::cerr << "Something is wrong - first daughter later than last"
		<< std::endl;
      return false;
    }

    // Now we set the daughter information on all mothers.  Next, we
    // must set the mother information on the daugthers, and recurse
    // down the tree
    for (auto p : out_range<VertexPtr>(v)) // v->particles_out()) 
      // Recursive call no this particle's terminating vertex 
      Convert(p->end_vertex(), a);

    return true;
  }
  /** 
   * For a given vertex set daughter information on mothers of that
   * vertex.  Note, after this member function has been called on a
   * vertex, the entry for this vertex will have the first and last
   * _mother_ index.  We will use that to fill in the mother
   * information.
   */
  bool FillInDaughters(VertexPtr v, std::pair<Int_t,Int_t>& fl, TClonesArray* a)
  {
    Int_t m1 = 1'000'000'000;
    Int_t m2 = -1;
    for (auto p : in_range<VertexPtr>(v)) { // ->particles_in()) {
      auto it = Map(p, a);
      if (it.first < 0) {
	std::cerr << "Mother particle still not made - this is weird" << std::endl;
	continue;
      }
      if (fl.first  >= 0) it.second->SetFirstDaughter(fl.first);
      if (fl.second >= 0) it.second->SetLastDaughter (fl.second);
      m1 = TMath::Min(it.first, m1);
      m2 = TMath::Max(it.first, m2);
    }
    fl.first  = m1;
    fl.second = m2;

    return true;
  }
  /** 
   * For a given vertex set mother information on daughters of that
   * vertex.  
   */
  bool FillInMothers(VertexPtr v, std::pair<Int_t,Int_t>& fl, TClonesArray* a)
  {
    if (fl.first < 0 and fl.second < 0) return true;

    for (auto p : out_range<VertexPtr>(v)) { // v->particles_out()) {
      auto it = Map(p, a);
      if (it.first < 0) {
	std::cerr << "Daughter particle not made - this is weird" << std::endl;
	continue;
      }
      if (fl.first  >= 0) it.second->SetFirstMother(fl.first);
      if (fl.second >= 0) it.second->SetLastMother (fl.second);
    }
    fSeenVertex[v] = std::make_pair(-1, -1);
    return true;
  }
    
};

//====================================================================
template <typename ParticlePtr,
	  typename VertexPtr,
	  typename FourVector,
	  typename Event>
struct GenEvent2Tree
{
  using Converter=GenEvent2Array<ParticlePtr,
				 VertexPtr,
				 FourVector,
				 Event>;
  
  TFile*         fOutput;
  TTree*         fTree;
  TClonesArray*  fArray;
  Event          fEvent;
  Converter      fConverter;
  
  GenEvent2Tree(const char* output="data.root")
    : fOutput(0),
      fTree(0),
      fArray(0),
      fEvent(),
      fConverter()
  {
    fOutput = TFile::Open(output,"RECREATE");
    fTree   = new TTree("T","T");
    fArray  = new TClonesArray("TParticle");
    fTree->Branch("Particles",&fArray);
  }
  virtual bool Read() = 0;
  bool Convert()
  {
    while (Read()) {
      fConverter.Convert(fEvent, fArray);
      fTree->Fill();
    }
    fTree->Write();
    fOutput->ls();
    fOutput->Close();
    return true;
  }
};



//====================================================================
# ifdef HEPMC3
#  include <HepMC3/ReaderAscii.h>
#  include <HepMC3/GenVertex.h>
#  include <HepMC3/GenParticle.h>
#  include <HepMC3/GenEvent.h>

struct HepMC2ROOT : public GenEvent2Tree<HepMC3::GenParticlePtr,
					 HepMC3::GenVertexPtr,
					 HepMC3::FourVector,
					 HepMC3::GenEvent>
{
  using Base=GenEvent2Tree<HepMC3::GenParticlePtr,
			   HepMC3::GenVertexPtr,
			   HepMC3::FourVector,
			   HepMC3::GenEvent>;
  HepMC3::ReaderAscii fReader;
    
  HepMC2ROOT(const char* in, const char* out)
    : Base(out),
      fReader(in)
  {}
  bool Read() override
  {
    if (!fReader.read_event(fEvent))
      return false;

    return !fReader.failed();
  }
};


# endif
# ifdef HEPMC2
#  include <HepMC/IO_GenEvent.h>
#  include <HepMC/GenVertex.h>
#  include <HepMC/GenParticle.h>
#  include <HepMC/GenEvent.h>

struct HepMC2ROOT : public GenEvent2Tree<HepMC::GenParticle*,
					 HepMC::GenVertex*,
					 HepMC::FourVector,
					 HepMC::GenEvent>
{
  using Base=GenEvent2Tree<HepMC::GenParticle*,
			   HepMC::GenVertex*,
			   HepMC::FourVector,
			   HepMC::GenEvent>;
  HepMC::IO_GenEvent fReader;
  
  HepMC2ROOT(const char* in, const char* out)
    : Base(out),
      fReader(in, std::ios::in)
  {}
  bool Read() override
  {
    return fReader.fill_next_event(&fEvent);
  }
};

# endif

//====================================================================
int main(int argc, char** argv)
{
  HepMC2ROOT c(argv[1], argv[2]);
  c.Convert();

  return 0;
}
  
  
#endif
```

ktf:
@TimoWilken can you please look into why the tests fail?

gconesab:
Hi @TimoWilken
any progress with this?
Thanks ... Gustavo

EVGEN/AliGenExtFile.cxx
amorsch: Hello, we could keep the two mothers. However in AliRoot, the mother->child relation is created on the fly from the child->mother relation and there is unfortunately no guarantee that this works for all MCs. For example in Pythia8 two mothers imo1, imo2 can mean: an interval from [imo1, imo2] or two separated mothers. Same for the daughter indices. AliStack assumes that the daughters are contiguous, which for MCs using two mother indices might not be always the case.

gconesab:
Hi @TimoWilken and @ktf,

Another ping to see if any of you could have a look.

Thanks ... Gustavo

chiarazampolli:
+1

ktf:
@preghenella there is some error in pythia, IIUC... Could it be the PR of yesterday?

ktf:
+test

chiarazampolli:
Hello @ktf , @TimoWilken , 

I cannot find the error in the build log, can you point me at it?

Chiara

TimoWilken:
@chiarazampolli the build fails because `finalcheck.sh` exits with an error status:
```
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: ++ ./finalcheck.sh
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: check File not finished
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: check File not finished
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: Pythia Warning
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: RunTracking not finished
2021-06-09@14:57:27:DEBUG:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: RunTracking not finished
2021-06-09@14:57:27:ERROR:AliRoot-guntest:AliRoot-guntest:build_AliRoot_release: Error while executing /mnt/mesos/sandbox/sandbox/aliroot-release/sw/SPECS/slc7_x86-64/AliRoot-guntest/v1-local1/build.sh on `alientest02'.
```

I'm guessing `finalcheck.sh` is [this file](https://github.com/alisw/AliRoot/blob/master/test/gun/finalcheck.sh), which searches for specific messages in various log files.

chiarazampolli:
Hi @TimoWilken , 

Thanks! So it looks like the sim part of this check did not work, and also the following steps then don't. But where are the logs? Is this what @ktf  referred to in a previous comment yesterday?

Chiara

gconesab:
Ciao Chiara,
I found this log 
https://ali-ci.cern.ch/alice-build-logs/alisw/AliRoot/1351/62c27c3820f935bb17af6bd7c628b9bce2fc1ba0/build_AliRoot_macos/fullLog.txt
the problem seems in the simulation. Apparently the test is using AliGenExtFile which is the one where the modifications are. I will try to reproduce locally.
Gustavo

gconesab:
The problem seems to be in the initialization of the fStack pointer in AliGenExtFile::Generate(), this pointer is not used in the master version of the class. The machinery of the modifications depends on the stack info and I do not see its initialization in any place Salvatore's commit. I do not know how to recover the stack, I will try to find out.

gconesab:
Hi all, I do understand why we had this comment from alibuild without approval, anyway can someone with powers do +test?

chiarazampolli:
+test

but i think we need someone else in addition...

gconesab:
Hi @shahor02, @jgrosseo, @qgp (Jochen Klein), @davidrohr, @maireiphc 
can any of you do +test?
Thanks

shahor02:
+test

shahor02:
+1

PR #1355, opened 2021-04-30
Sync AliGPU from O2

davidrohr:


davidrohr:
+test

davidrohr:
@TimoWilken @MichaelLettrich : The MacCI seems to fail here without showing a real error.

TimoWilken:
It looks to me like the error mentioned in the HTML log (`error: 'auto' return without trailing return type; deduced return types are a C++14 extension`) is the one that killed the build.

On a separate note, I'm not sure why the AliRoot log isn't last in the plain-text log -- it is the last thing that was built, and the build error occurred during AliRoot compilation.

davidrohr:
ah, ok, thanks. I was only looking at the last log. Indeed, the sorting is wrong.

davidrohr:
+test

davidrohr:
+test

davidrohr:
+test

davidrohr:
+1

PR #1358, opened 2021-05-26
protection against -1 index

loizides:
, which means that the particle made it outside the
defined geometry, ie gGeoManager->IsOutside() == kTRUE


amorsch:
Discussed with Constantin and OK.
+1

PR #1367, opened 2021-07-20
Fix Omega_c lifetime in PYTHIA8 according to PDG2020

zhujh126:
Fix Omega_c lifetime in PYTHIA8 according to PDG2020


chiarazampolli:
+1

fprino:
This modification is needed for a MC production that was approved today in the PB.
The modified value of ctau is taken from the latest measurement of omega_c lifetime by LHCb reported in the PDG.

PR #1387, opened 2021-08-12
Adds cout statement to verify that L1 noise is being set properly.

moliver813:


EMCAL/EMCALsim/AliEMCALTriggerElectronics.cxx
```diff
@@ -39,6 +39,11 @@
 #include <TRandom.h>
 #include <TVector2.h>
 
+#include <iostream>
+#include <cstdlib>
+
+using namespace std;
```
mfasDa, line 45: NEVER do `using namespace std` - this is potentially dangerous!

EMCAL/EMCALsim/AliEMCALTriggerElectronics.cxx
```diff
@@ -76,6 +81,8 @@ fGeometry(0)
     if (AliDebugLevel()) fSimParam->Print("");
   }
 
+  cout << "fL1ADCNoise =" << fL1ADCNoise << endl;
```
mfasDa, line 84: change to std::cout and std::endl and drop `using namespace std`.

mfasDa (changes requested):
Code must not have `using namespace std`

mfasDa:
+1

mfasDa:
Though I would have used something like

``` c++
std::cout << "EMCAL Trigger Electronics: Using ADC threshold " << fL1ADCNoise << std::endl;
```

That would been easier to digest. However I don't want to delay the PR any further.

PR #1391, opened 2021-08-26
cmake:: fix doxygen finding on f34

adriansev:
on cmake 3.20.5 there are problems if doxygen is already found. Just invalidate the already detected doxygen executable.
solution from https://gitlab.kitware.com/cmake/cmake/-/issues/18708


adriansev:
@ktf this is wierd, at some point the checks were green and now i see this unknown problem without any logs .. for reference the same change https://github.com/alisw/AliPhysics/pull/18384 is green now...

adriansev:
@ktf ping? :) sorry for pestering you

PR #1394, opened 2021-12-09
Enable VMC standalone

benedikt-voelkel:


benedikt-voelkel:
@sawenzel @ihrivnac @amorsch 

So that AliRoot can still be built with this https://github.com/alisw/alidist/pull/3640

benedikt-voelkel:
Hello @amorsch can we test this?

sawenzel:
+test

amorsch:
+test

benedikt-voelkel:
That looks a bit strange to me. I see no problems with the ROOT `--defaults release` built locally and actually it goes through including AliRoot including all tests
Is that a problem on the machine?
@TimoWilken maybe you can see something?

benedikt-voelkel:
@pzhristov can we merge this as well? That should be the last bit before we can tackle alidist for real this time

PR #1405, opened 2022-02-26
Rescale ITS material density for budget studies

mario6829:
This commit introduces the possibility of varying the density of some materials used in the implementation of the SPD and SDD detectors, in order to perform material budget studies. By default no change takes places, but at run time the user has the possibility to multiply the density of some chosen materials by a scale factor.


chiarazampolli:
+1

PR #1406, opened 2022-02-28
Scale mat density

chiarazampolli:
@mario6829 , @noferini 
If you agree, I would merge. This is the tag for pp.


noferini:
It's fine with me

PR #1403, opened 2022-02-14
Fix memory leak in AliAODv0 constructor

fprino:
I spotted a leak in AliAODv0, which was causing a memory increase in the tests for conversions of HF candidates from ESD to AO2D.

In the constructor of AliAODv0 (which inherits from AliAODRecoDecay) we have:

AliAODv0::AliAODv0(AliAODVertex* rAODVertex, Double_t rDcaV0Daughters, Double_t rDcaV0ToPrimVertex,
        const Double_t *rMomPos, const Double_t *rMomNeg, Double_t *rDcaDaughterToPrimVertex) :
   AliAODRecoDecay(rAODVertex,2,0,rDcaDaughterToPrimVertex),
   fDcaV0ToPrimVertex(rDcaV0ToPrimVertex),
   fOnFlyStatus(kFALSE)
{
[...]
   fDCA = new Double_t[fNDCA];

[...]
   fPx = new Double_t[GetNProngs()];
   fPy = new Double_t[GetNProngs()];
   fPz = new Double_t[GetNProngs()];

[...]
}

The leak is caused by the fact that the constructor that is invoked for AliAODRecoDecay already creates fPx, fPy and fPz:

AliAODRecoDecay::AliAODRecoDecay(AliAODVertex *vtx2,Int_t nprongs,
                  Short_t charge,
                  Double_t *d0) :
   AliVTrack(),
   fSecondaryVtx(vtx2),
   fOwnSecondaryVtx(0x0),
   fCharge(charge),
   fNProngs(nprongs), fNDCA(0), fNPID(0),
   fPx(0x0), fPy(0x0), fPz(0x0),
   fd0(0x0),
   fDCA(0x0),
   fPID(0x0)
{
   /// Constructor with AliAODVertex for decay vertex and without prongs momenta

   fPx = new Double32_t[GetNProngs()];
   fPy = new Double32_t[GetNProngs()];
   fPz = new Double32_t[GetNProngs()];
   fd0 = new Double32_t[GetNProngs()];
   for(Int_t i=0; i<GetNProngs(); i++){
     fPx[i] = 0;;
     fPy[i] = 0.;
     fPz[i] = 0.;
     fd0[i] = d0[i];
   }
}

This PR applies a simple fix.

PR #1411, opened 2022-02-28
Fix memory leak in AliAODv0 constructor

chiarazampolli:
@fprino , @noferini .
This is the tag for MC anchored to 18q/r pass3.


chiarazampolli:
Hi @TimoWilken , @ktf , 

This was put to master by mistake, it was in fact already there from:
https://github.com/alisw/AliRoot/pull/1403/files

is it correct that the "Files changed" tab reports something?
Chiara

PR #1421, opened 2022-03-21
Change include to allow for inclusion of AliPythia8 in analysis task

ddobrigk:


ddobrigk:
For reference: this is required to have an AliPhysics analysis compile while using AliPythia8.h, which in turn includes AliDecayerPythia8. It's important to note that invoking the AliPythia8 singleton will not be particularly useful in regular analyses over already processed MC, but it will be useful in the MCgen trains as those generate each and every event before processing the analysis wagons; this, in turn, means that the relevant information is still in memory when the analysis is running. 

It would be really great if we could move this to AliRoot soon, considering also the triviality of the change. Could someone please help out? Thank you very much!

pzhristov:
+test

PR #1427, opened 2022-04-06
Fix for hypernuclei masses

jditzelnew:
The masses of the hyper hydrogen4 and the hyper helium4 need to be corrected since in the old implementation only the lambda separation energy of 2MeV/c2 was substracted from the sum of the constituents. This is incorrect since also the binding energy of the remaining nucleus (triton and he3 respectively) needs to be taken into account. Furthermore there is the existing of an excited state for both particles (see A. Gal, E.V. Hungerford and D.J. Millener, „Strangeness in nuclear physics“, Rev. Mod. Phys. 88 (2016) 3, 035004, arXiv:1605.00557 [nucl-th]) which is implemented now with the correct mass after the (very short lived) EM transition to the ground state.


shahor02:
+1

PR #1428, opened 2022-05-16
Sync AliGPU from O2

davidrohr:


davidrohr:
+test

davidrohr:
+1

davidrohr:
@ktf @pzhristov @TimoWilken : The MacCI always fails at Geant3. Is this a known problem?

davidrohr:
+1

pzhristov:
It seems root-config --has-vmc returns "yes", which is not expected after we moved to separate VMC distribution. Most probably we have to remove the line  ${BUILD_VMC_INTERNAL:+-Dvmc=ON} from alidist/root.sh. Anyway, I tested the changes locally on Mac M1 and everything is fine.

PR #1434, opened 2022-07-14
Results-based strange particle pt shapes for pp5 to be used in GenParam

bianchil:


bianchil:
Dear all,
as the commit is necessary for a MC request that we are eager to have, I was wondering if there is anything problematic with this or if I can bug you for a +1 so that we can proceed. Thank you very much in advance!
Putting @alcaliva in cc for info

chiarazampolli:
+1

chiarazampolli:
I am not an expert of this code, but it seems fine to me. Approving to unblock MC.

TimoWilken:
Hi @chiarazampolli, see alisw/alidist#4356 (now merged). I've restarted the check.

PR #1438, opened 2022-09-26
Adapting STARlight to ROOT6

grundd:
No major changes were made, previously the STARlight generator was not compatible with ROOT6. Errors were solved by renaming the variables "_Z" and "_A" in nucleus.h without using underscores and commenting out a constant "kconv" in AliDecayerSLEvtGen.h


grundd:
The changes were consulted with Ionut Arsene, Joakim Nystrand and Michal Broz. @iarsene @mbroz84

PR #1441, opened 2022-10-14
lego_train.C: Only load libraries manually under ROOT5

TimoWilken:
This should fix a segfault encountered when libVMC or libSTEERBase were loaded manually twice.

A test train with these changes works on Alimonitor and on the Grid.


pzhristov:
@TimoWilken Please add #include <RVersion.h>, otherwise ROOT_VERSION is not defined.

TimoWilken:
Done, @pzhristov!

pzhristov:
+1

PR #1443, opened 2022-10-21
Explicitly specify pythia6 dependency for AliPythia6

TimoWilken:
Previously, analysis trains relied on the pythia6 library being pre-loaded, but that pre-loading is a ROOT5-ism. Instead, make sure libAliPythia6 is linked properly against pythia6.

This should solve the `libAliPythia6.so: undefined symbol: pyqpar_` error seen by analysers.

This seems to fix the problem in the one instance I reproduced locally based on the run at:
http://alitrain.cern.ch/train-workdir/PWGGA/GA_pp_MC/3980_20221018-2243/test/__BASELINE__

While I did not reproduce the full run with all *.root files, I managed to reproduce the crash (without this patch), which did not occur with this patch applied.


TimoWilken:
The Linux check failed due to a known problem unrelated to this PR; merging.

PR #1448, opened 2022-11-30
fix issue for the direct gamma generation in triggered pythia

iarsene:


gconesab:
Hi Ionut, I do not see why this is needed. fPhotonInCalo is not used, I guess I left it for backward compatibility.

iarsene:
Hi Gustavo, we use it now in AliGenPythiaFOCAL, which is in the FOCAL package. This seems to cure a problem we were having in generating direct isolated gamma.

chiarazampolli:
+1

TimoWilken:
Errors unrelated; merged.

PR #1450, opened 2023-01-10
AliDecayerPythia: re-write of forced EtaPrime decays

fjonasALICE:
- fix for eta prime decays to force eta to decay to gamma gamma
- separate setting for EtaPrime to GammaGamma decay


fjonasALICE:
this change is needed, as it was discovered that simulations with forced eta prime to pipieta decay do not offer sufficient statistics for physics analysis in pp13TeV, since the BR for eta->gammagamma was not accounted for. This is now fixed and the decay eta->gammagamma is forced as well.

In addition the previous implementation for etaprime to gammagamma decays was handled in an odd way and decided by a random number. We propose splitting both decay channels into separate flags.

FriederikeBock:
@sawenzel or @pzhristov would you mind approving please. We'll need to back port this change to one of the older aliroot versions to redo the orginal eta' production

PR #1453, opened 2023-01-17
Get the generated vertex position of trigger event in case of pileup

fprino:
This modification to AliMCEvent::GetPrimaryVertex allows one to retrieve the generated vertex position for the trigger collision in case of AODs of MC productions with generated pileup.

This method was already returning the position of the trigger event for ESDs.
This is due to the fact that the vertex position of the trigger event is properly set in the "highest level" header of the cocktail  generator and it is thus correctly retrieved in AliMCEvent::GetPrimaryVertex.

However, this does not hold in the case of AODs, for which the "highest level" cocktail header is not made available in AliAODMCHeader (and therefore in AliMCEvent), due to the different structure of the event headers of the cocktail generator it was returning (0,0,0).
Nevertheless, the manipulations done in AliAODHandler::StoreMCParticles store the vertex position (taken from the highest level cocktail header, hence the same source used for ESDs) into the data member fVertex of AliAODMCHeader.
With this modification the proper vertex position in case of AODs with pileup generation is retrieved by using AliAODMCHeader::GetVtxX,Y,Z.

### DIFF
--- a/STEER/STEERBase/AliMCEvent.cxx
+++ b/STEER/STEERBase/AliMCEvent.cxx
@@ -919,7 +919,18 @@ const AliVVertex * AliMCEvent::GetPrimaryVertex() const
 {
     // Create a MCVertex object from the MCHeader information
     TArrayF v;
-    GenEventHeader()->PrimaryVertex(v) ;
+    AliGenEventHeader* gh =  GenEventHeader();
+    gh->PrimaryVertex(v) ;
+    
+    // Special treatment for AODs from productions with pileup generation
+    TString gname=gh->GetName();
+    gname.ToLower();
+    if(fAODMCHeader && gname.Contains("pileup") && gh->InheritsFrom(AliGenCocktailEventHeader::Class())){
+      v[0]=fAODMCHeader->GetVtxX();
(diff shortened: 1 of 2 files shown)


chiarazampolli:
+1

chiarazampolli:
Hello @TimoWilken , 
Do you know where the error here comes from?
Cheers,
Chiara

TimoWilken:
Hi @chiarazampolli, sorry about that, I had a fix installed on one of the Mac builders that got out of sync with recent alidist changes. I've reset that now.

The `build/AliRoot/o2` check has been broken for a while I think, though at least it manages to compile AliRoot fine, and it gets most of the way through the guntest...

sstiefel19:
Hi @TimoWilken, are you still following this? (I see the builds are restarted regularly but not sure if this automatically of you..)

TimoWilken:
Hi @sstiefel19, sorry about that, the Mac one is fixed now. The "o2" check manages to compile AliRoot but falls over during the guntest as with other PRs, so I'll merge this.

sstiefel19:
great, thanks!

PR #1464, opened 2023-03-24
Introduce f0_1500 resonance

BongHwi:
Related discussion: [ALIROOT-8826](https://alice.its.cern.ch/jira/browse/ALIROOT-8826?focusedCommentId=297923&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-297923)

We'd like to implement f0_1500 resonance in the particle list and decay table.

But I'm not sure about the parameters that I set here.
Especially, I don't have any idea about the arguments of `TVirtualMC::GetMC()->DefineParticle`:
`TVirtualMC::GetMC()->DefineParticle(9030221, "f0_1500", kPTNeutron, 1.505, 0.0, 1.0e-23, "Hadron", 0.109, **0, 1, 1, 0, 0, 1, 0, 0, kTRUE);**`

it would be great if you can help us with this value.


BongHwi:
Quick update: DefineParticle arguments are correct including the spin parities (same with f0_980).
but I'm not sure about the lifetime...

PR #1470, opened 2023-04-22
Add decay table for non-prompt psi(2S)

JinjooSeo:
Add decay table for non-prompt psi(2S) in dielectron channel


gconesab:
Dear @JinjooSeo 
is this needed for an existing production request, can you point me to it? Thanks ... Gustavo

PR #1472, opened 2023-05-05
Updating the mass of exotic bound states

BongHwi:
The new value is from the 3-body decay channel: anti-p anti-p K^+
The total mass of the decay products is 0.938+0.938+0.493 = 2.37 GeV.


BongHwi:
Could you review this? @pzhristov

PR #1479, opened 2023-06-01
Fix PDG charges (after 20 years)

vkucera:
Fixing wrong charges added in commits:
https://github.com/alisw/AliRoot/commit/8ff209d4372162d79b02c86dbfb4c41fc8595177
https://github.com/alisw/AliRoot/commit/d3df3965d473189007440b8a7e4d36f2905c8bc3
partially because of wrong ROOT documentation,
which was fixed later in https://github.com/root-project/root/commit/6a82edd615f1284f2de238bab195a
but not propagated to AliRoot.

### DIFF
--- a/STEER/STEERBase/AliPDG.cxx
+++ b/STEER/STEERBase/AliPDG.cxx
@@ -26,11 +26,8 @@
 
 ClassImp(AliPDG)
 
-
-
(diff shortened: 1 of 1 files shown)


pzhristov:
+1

PR #1486, opened 2023-06-28
Fix PDG charges (after 20 years) and others to V5 09 54 patches

gconesab:


chiarazampolli:
Same as in https://github.com/alisw/AliRoot/pull/1485: @TimoWilken , could you merge?

PR #1488, opened 2023-07-14
PWG DQ: adding selection on the child for powheg processes for AliGenPythia

michelepennisi:
Adding selection on the child for powheg processes for AliGenPythia


chiarazampolli:
+1

PR #1502, opened 2023-10-24
Add width of Ds resonances

gconesab:
Needed for ALIROOT-8847


gconesab:
Hi @TimoWilken can you merge and re-tag? If possible including the Geant4 last recipes.

PR #1508, opened 2023-12-13
Fix a different version of Pythia8

ktf:


ktf:
Same fix as https://github.com/alisw/AliRoot/commit/2bbfba44fad7ba639548b3d3061ea0be85fc1f19, just for a different version of pythia8 (which is the one actually used).

PR #1511, opened 2024-02-01
Update AliDecayerPythia8.cxx

ddobrigk:


ddobrigk:
This suggested change is such that the code now follows the revised ALICE primary particle definition regarding what to decay and what not to decay, avoiding the risk of forgetting to disable / enable any particles manually

PR #1514, opened 2024-03-22
Bugfix: AliGenSexaquarkReaction

aborquez:
two bug fixes:
- default call of SetPhiRange is now in correct units (degrees instead of radians)
- momentum distributions for Fermi motion are now flat


gconesab:
Hello, it seems a rather trivial fix, can anyone approve it? @chiarazampolli @shahor02

chiarazampolli:
+1

